PR #1, opened 2020-05-20
Use loop label in InnerRingUpdate call handler

cthulhu-rider:


AlexVanin:
`continue loop` support came with neo-go v0.74.0: https://github.com/nspcc-dev/neo-go/pull/678

PR #3, opened 2020-05-26
Do not store multiaddress of inner ring nodes

AlexVanin:
With neo:morph environment, there will be no direct communication between inner ring nodes and storage nodes. neo:morph smart-contracts will identify inner ring nodes by their signatures.


AlexVanin:
I will rebase PR after #2

PR #4, opened 2020-05-28
Collect "InnerRingUpdate" calls from inner ring nodes 

AlexVanin:
Inner ring nodes do not collect signatures for the cheque now. Instead they invoke "InnerRingUpdate" method and smart-contract checks if method was called from inner ring node. Then it accepts cheque if there were 2/3n+1 invokes.

### DIFF
--- a/neofs_contract.go
+++ b/neofs_contract.go
@@ -8,21 +8,29 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/interop/util"
 )
 
-type node struct {
-	pub []byte
-}
+type (
+	ballot struct {
+		id []byte
+		n  int
+	}
 
-type check struct {
-	id     []byte
-	height []byte
-}
+	node struct {
+		pub []byte
+	}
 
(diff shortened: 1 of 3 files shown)


neofs_contract.go
```diff
@@ -405,3 +422,61 @@ func verifySignatures(irList []node, data []byte, message []byte, offset int) bo
...
+	data := runtime.Serialize(newCandidates)
+	storage.Put(ctx, voteKey, data)
+
+	return found
+}
+
+func removeVotes(ctx storage.Context, id []byte) {
+	var (
+		newCandidates = []ballot{}
+		candidates    = getSerialized(ctx, voteKey).([]ballot)
```
cthulhu-rider, line 470: Can we just cut an element from `candidates` slice below instead of filling `newCandidates` slice?

AlexVanin: No, for now we can't: https://github.com/nspcc-dev/neo-go/issues/999

PR #13, opened 2020-07-16
Add neofs install configuration

AlexVanin:
Rebase and review after #12

### DIFF
--- a/neofs_contract.go
+++ b/neofs_contract.go
@@ -6,6 +6,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/interop/contract"
 	"github.com/nspcc-dev/neo-go/pkg/interop/crypto"
 	"github.com/nspcc-dev/neo-go/pkg/interop/engine"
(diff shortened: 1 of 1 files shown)


AlexVanin:
Rebased and ready for the review.

neofs_contract.go
```diff
@@ -369,6 +381,91 @@ func Main(op string, args []interface{}) interface{} {
...
+
+		return true
+	case "ListConfig":
+		var config []record
+
+		it := storage.Find(ctx, configPrefix)
+		for iterator.Next(it) {
+			key := iterator.Key(it).([]byte)
+			val := iterator.Value(it).([]byte)
+			r := record{key: key[6:], val: val}
```
cthulhu-rider, line 441: @AlexVanin, is `6` the length of `configPrefix`? If so, it would be convenient to clarify this through `len(configPrefix)` or contst.

AlexVanin: Yes, it is. I am agree with you, length will be more convenient there.

PR #20, opened 2020-11-16
Add committee the ability to vote for new inner ring list

alexvanin:
Now committee can change inner ring list by collecting committee votes. However new inner ring list proposed by committee should fit the same criteria: it should contain keys from candidate list and at most 1\3 of new nodes.

### DIFF
--- a/neofs/neofs_contract.go
+++ b/neofs/neofs_contract.go
@@ -68,6 +68,9 @@ const (
 	// native gas token script hash
 	tokenHash = "\xbc\xaf\x41\xd6\x84\xc7\xd4\xad\x6e\xe0\xd9\x9d\xa9\x70\x7b\x9d\x1f\x0c\x8e\x66"
 
+	// native neo token script hash
+	neoHash = "\x25\x05\x9e\xcb\x48\x78\xd3\xa8\x75\xf9\x1c\x51\xce\xde\xd3\x30\xd4\x57\x5f\xde"
+
 	defaultCandidateFee   = 100 * 1_0000_0000 // 100 Fixed8 Gas
 	candidateFeeConfigKey = "InnerRingCandidateFee"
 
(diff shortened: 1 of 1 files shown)


alexvanin:
Checked on a single node environment with stand by committee.

neofs/neofs_contract.go
```diff
@@ -520,6 +531,17 @@ func innerRingInvoker(ir []node) []byte {
 	return nil
 }
 
+func committeeInvoker(cmt [][]byte) []byte {
+	for i := 0; i < len(cmt); i++ {
+		node := cmt[i]
+		if runtime.CheckWitness(node) {
```
roman-khimov, line 537: This means that each committee member has to invoke `innerRingUpdate` and I'm not sure this is the way it's supposed to work. At the same time you probably can't really check for committee multisignature.

neofs/neofs_contract.go
```diff
@@ -338,9 +343,15 @@ func InnerRingUpdate(chequeID []byte, args [][]byte) bool {
...
 
-	irKey := innerRingInvoker(irList)
-	if len(irKey) == 0 {
-		panic("innerRingUpdate: invoked by non inner ring node")
+	voteKey := innerRingInvoker(irList)
+	if len(voteKey) == 0 {
+		committee := getCommitteeKeys()
+		threshold = len(committee)/3*2 + 1
+
+		voteKey = committeeInvoker(committee)
```
roman-khimov, line 351: And this treats IR votes and committee votes equal, I'm not sure about that either.

roman-khimov (commented):
I think the intention for committee-based voting was to use Designate contract and add a role there for IR nodes. Although moving IR nodes list there will also require some kind of synchronization between two contracts and Designate generates no events at the moment.

alexvanin:
I am moving this PR to draft, since there will be changes in new committee/innerring communication scheme. It will require support of committee signagure check in contract.

alexvanin:
I close this PR because there will be updated NeoFS governance scheme that has nothing to do with this PR (#32)

PR #23, opened 2020-12-09
Support NEP-17 and other preview4 features

### LINKED ISSUE #21: Deploy alphabet contracts from single binary
After neo-project/neo#2044 we might be able to deploy the same binary contract as different contracts with different addresses. This way we can get rid of alphabet template in repository. In this case each alphabet contract should be initialized once with index for committee voting.

_Originally posted by @roman-khimov in https://github.com/nspcc-dev/neofs-contract/pull/19#discussion_r527828658_

alexvanin:
This pull requests provides new preview4 compatible features.

- [x] Balance contract is a NEP-17 compatible contract
- [x] NeoFS contract supports deposits via `onPayment` callback
- [x] ~~More explicit types such as `interop.Hash160`, `interop.Hash256` and `interop.PublicKey` in method signatures to improve contract readability~~ (will be done in separate issue #25)
- [x] Single alphabet contract instead of template (Closes #21)


alexvanin:
Ready for review. Closes #21

PR #45, opened 2021-02-02
Share common code between contract

### LINKED ISSUE #42: Reorganize common contract code in util package
There are a lot of common structures and functions in contracts such as:
- `ballot struct`,
- `vote()`,
- `removeVote()`,
- `invokeID()`,
- `innerRingInvoker()`,
- ...
- many more.

They can be placed in separate package and it should be compiled with neo-go. This will heavily reduce complexity and size of the contracts.

cthulhu-rider:
Closes #42.

- [x] `type ballot`
- [x] `func vote`
- [x] `func removeVote`
- [x] `func invokeID`
- [x] `func innerRingInvoker`
- [x] `func setSerialized`
- [x] `func bytesEqual`


common/ballot.go
```diff
@@ -0,0 +1,12 @@
+package common
+
+type Ballot struct {
+	// ID of the voting decision.
+	ID []byte
+
+	// Number already voted inner ring nodes.
```
alexvanin, line 7: `Public keys of already voted inner ring nodes` will be more accurate.

cthulhu-rider: Fixed

common/ballot.go
```diff
@@ -0,0 +1,12 @@
+package common
+
+type Ballot struct {
```
alexvanin, line 3: I think ballot type can be defined in `vote.go`, it has not use outside of voting.

common/invoke.go
```diff
@@ -0,0 +1,31 @@
+package common
+
+import (
+	"github.com/nspcc-dev/neo-go/pkg/interop/crypto"
+	"github.com/nspcc-dev/neo-go/pkg/interop/runtime"
+)
+
+func InvokeID(args []interface{}, prefix []byte) []byte {
```
alexvanin, line 8: `InvokeID` used to identify txs and collect votes, so it better to move to `vote.go`.

common/invoke.go
```diff
@@ -0,0 +1,31 @@
...
+	for i := range args {
+		arg := args[i].([]byte)
+		prefix = append(prefix, arg...)
+	}
+
+	return crypto.SHA256(prefix)
+}
+
+type IRNode struct {
```
alexvanin, line 17: Let's move inner ring list related code to `ir.go`: both `IRNode` struct and `InnerRingInvoker` function.

alexvanin (approved):

PR #54, opened 2021-03-04
Feature/use explicit types for hashes

### LINKED ISSUE #25: Use explicit types for public keys and hashes
neo-go has [explicit type declarations](https://github.com/nspcc-dev/neo-go/blob/master/pkg/interop/types.go) for public keys, script hashes and SHA256 hashes. Even though they all processed as `[]byte` and don't guarantee length or correctness, it will be better to use them. This  is pure cosmetic change, however it makes easier to understand source code.

Possible enhancements:
- use `interop.Hash160` in all `Init` functions,
- use `interop.PublicKey` in inner ring related structures,
- use `interop.Hash256` in vote ID generation.

carpawell:
Closes #25.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.manifest.json
 config.json
 /vendor/
+.idea
--- a/alphabet/alphabet_contract.go
+++ b/alphabet/alphabet_contract.go
@@ -80,7 +80,7 @@ func irList() []common.IRNode {
 }
 
 func currentEpoch() int {
-	netmapContractAddr := storage.Get(ctx, netmapKey).([]byte)
+	netmapContractAddr := storage.Get(ctx, netmapKey).(interop.Hash160)
 	return contract.Call(netmapContractAddr, "epoch", contract.ReadOnly).(int)
 }
 
@@ -143,7 +143,7 @@ func Emit() bool {
 	return true
 }
 
-func Vote(epoch int, candidates [][]byte) {
+func Vote(epoch int, candidates []interop.PublicKey) {
 	index := index()
 	name := name()
 
--- a/container/container_contract.go
+++ b/container/container_contract.go
@@ -111,7 +111,7 @@ func Put(container []byte, signature interop.Signature, publicKey interop.Public
 		return true
 	}
 
-	from := walletToScripHash(ownerID)
+	from := walletToScriptHash(ownerID)
 	netmapContractAddr := storage.Get(ctx, netmapContractKey).([]byte)
 	balanceContractAddr := storage.Get(ctx, balanceContractKey).([]byte)
 	containerFee := contract.Call(netmapContractAddr, "config", contract.ReadOnly, containerFeeKey).(int)
@@ -437,7 +437,7 @@ func getEACL(ctx storage.Context, cid []byte) extendedACL {
 	return extendedACL{val: []byte{}, sig: interop.Signature{}, pub: interop.PublicKey{}}
 }
 
-func walletToScripHash(wallet []byte) []byte {
+func walletToScriptHash(wallet []byte) []byte {
 	return wallet[1 : len(wallet)-4]
 }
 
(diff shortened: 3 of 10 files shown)


neofsid/neofsid_contract.go
```diff
@@ -29,7 +29,7 @@ func init() {
 	ctx = storage.GetContext()
 }
 
-func Init(owner interop.Hash160, addrNetmap, addrContainer []byte) {
+func Init(owner interop.Hash160, addrNetmap, addrContainer interop.Hash160) {
```
cthulhu-rider, line 32:
```suggestion
func Init(owner, addrNetmap, addrContainer interop.Hash160) {
```

alphabet/alphabet_contract.go
```diff
@@ -31,7 +31,7 @@ func init() {
 // OnNEP17Payment is a callback for NEP-17 compatible native GAS and NEO contracts.
 func OnNEP17Payment(from interop.Hash160, amount int, data interface{}) {
 	caller := runtime.GetCallingScriptHash()
-	if !common.BytesEqual(caller, []byte(gas.Hash)) && !common.BytesEqual(caller, []byte(neo.Hash)) {
+	if !common.BytesEqual(caller, interop.Hash160(gas.Hash)) && !common.BytesEqual(caller, interop.Hash160(neo.Hash)) {
```
cthulhu-rider, line 34: Do we really need to cast to `interop.Hash160`? `common.BytesEqual` accepts `[]byte`

alexvanin: `gas.Hash` and `neo.Hash` are string constants, that's why we have to cast it either to `[]byte` or `interop.Hash160` in this case, since it is just an alias.

container/container_contract.go
```diff
@@ -434,14 +434,14 @@ func getEACL(ctx storage.Context, cid []byte) extendedACL {
...
+	return extendedACL{val: []byte{}, sig: interop.Signature{}, pub: interop.PublicKey{}}
 }
 
-func walletToScripHash(wallet []byte) []byte {
+func walletToScriptHash(wallet []byte) []byte {
 	return wallet[1 : len(wallet)-4]
 }
 
-func verifySignature(msg, sig []byte, keys [][]byte) bool {
+func verifySignature(msg, sig interop.Signature, keys []interop.PublicKey) bool {
```
alexvanin, line 444:
```suggestion
func verifySignature(msg []byte, sig interop.Signature, keys []interop.PublicKey) bool {
```

proxy/proxy_contract.go
```diff
@@ -23,12 +23,12 @@ func init() {
 
 func OnNEP17Payment(from interop.Hash160, amount int, data interface{}) {
 	caller := runtime.GetCallingScriptHash()
-	if !common.BytesEqual(caller, []byte(gas.Hash)) {
+	if !common.BytesEqual(caller, interop.Hash160(gas.Hash)) {
 		panic("onNEP17Payment: alphabet contract accepts GAS only")
```
alexvanin, line 27: Let's fix this panic message typo in this PR with extra commit: `onNEP17Payment: proxy contract accepts GAS only`.

alexvanin (changes requested):
Write commit message summaries starting with uppercase letter. Nothing wrong with lowercase but stick with the chosen commit message style. And try to keep them short, details could be described in commit description.

proxy/proxy_contract.go
carpawell: attached additional commit to this PR

container/container_contract.go
carpawell: my bad, fixed

alphabet/alphabet_contract.go
carpawell: so would it be better for readability to revert back or to keep changes?

.gitignore
```diff
@@ -3,3 +3,4 @@
 *.manifest.json
 config.json
 /vendor/
+.idea
```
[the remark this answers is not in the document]
carpawell: after discuss decided to keep that change for potential external contributors

neofsid/neofsid_contract.go
carpawell: my bad, fixed

alphabet/alphabet_contract.go
cthulhu-rider: `interop.Hash160` type does not provide API to cast it to `[]byte`, so I'd recommend to keep `string -> []byte` cast.

alexvanin (approved):

alphabet/alphabet_contract.go
carpawell: ok, rolled back

cthulhu-rider (approved):

PR #57, opened 2021-03-10
Feature/use readonly context

### LINKED ISSUE #37: Use read only storage context where it possible
There are kinda overwhelming usage of r/w context in the contracts. Now it is possible to use `storage.GetReadOnlyContext()` getting r/o context and use it in methods like `netmap.Snapshot`, etc.

carpawell:
Closes #37.

### DIFF
--- a/audit/audit_contract.go
+++ b/audit/audit_contract.go
@@ -40,13 +40,9 @@ const (
 	netmapContractKey = "netmapScriptHash"
 )
 
-var ctx storage.Context
-
-func init() {
-	ctx = storage.GetContext()
-}
-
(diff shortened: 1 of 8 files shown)


audit/audit_contract.go
```diff
@@ -119,12 +121,14 @@ func ListByCID(epoch int, cid []byte) [][]byte {
 	var buf interface{} = epoch
 
 	prefix := append(buf.([]byte), cid...)
+	ctx := storage.GetReadOnlyContext()
```
alexvanin, line 124: Let's move it up to the top of the function as it everywhere else.

carpawell: moved

PR #61, opened 2021-03-17
Run cleanup routines on new epoch

### LINKED ISSUE #53: Inner ring should invoke contract clean up methods every epoch
## Expected Behavior
Every epoch some data on sidechain become invalid and inner ring should invoke corresponding clean up methods. 

### Withdraw locks
Withdraw operation locks assets on account for some time and if mainnet transfer failed, then lock assets are returned back. Lock account valid for some epochs and inner ring should invoke [NewEpoch](https://github.com/nspcc-dev/neofs-contract/blob/9cdfcd438ed3b7e12bbf65bed731bd33097f468f/balance/balance_contract.go#L182) method so the contract will drop invalid lock accounts.

### Container estimation
There is no point to store all container estimations in side chain snapshots. These are also valid for couple epochs and then can be dropped. Inner ring again should invoke [ProcessEpoch](https://github.com/nspcc-dev/neofs-contract/blob/9cdfcd438ed3b7e12bbf65bed731bd33097f468f/container/container_contract.go#L338) method to this.

## Current Behavior
Right now the only cleanup job inner does is to remove unresponsive storage nodes from network map (only if it is enabled by `NEOFS_IR_NETMAP_CLEANER_ENABLED` option).

## Possible Solutions
1. Inner ring invokes `container.ProcessEpoch()` and `balance.NewEpoch()` methods after receiving new epoch notification.
2. Contracts can invoke each other directly, e.g. netmap contract invokes both container and balance contracts. 

Example(1):
```go
func (np *Processor) processNewEpoch(epoch uint64) {
	. . .
	np.netmapSnapshot.update(snapshot, epoch)
	np.handleCleanupTick(netmapCleanupTick{epoch: epoch})
	np.handleNewAudit(audit.NewAuditStartEvent(epoch))
	np.handleAuditSettlements(settlement.NewAuditEvent(epoch))
	
	// cleanup invocations
	invoke.NewEpochCleanup(np.morphClient, np.balanceContract, epoch)
	invoke.NewEpochCleanup(np.morphClient, np.containerContract, epoch)
}
```

Pros (2): 
- no inner ring logic involved.

Cons (2): 
- large cost of `netmap.NewEpoch` invocation and corresponding issues with nspcc-dev/neofs-node#47 in this case.

## Additionally
- [x] Sync method signatures for such new epoch cleanup routines between contracts.

alexvanin:
Closes #53 

Additionally: 
- remove unused code from neofsid contract after notary update,
- use `interop.Hash160` type for contract addresses fetched from storage in container contract.

### DIFF
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -39,6 +39,10 @@ const (
 	snapshot0Key  = "snapshotCurrent"
 	snapshot1Key  = "snapshotPrevious"
 	snapshotEpoch = "snapshotEpoch"
+
+	containerContractKey = "containerScriptHash"
+	balanceContractKey   = "balanceScriptHash"
+	cleanupEpochMethod   = "newEpoch"
 )
 
 const (
@@ -53,11 +57,15 @@ var (
 
 // Init function sets up initial list of inner ring public keys and should
 // be invoked once at neofs infrastructure setup.
-func Init(owner interop.Hash160, keys []interop.PublicKey) {
+func Init(owner, addrBalance, addrContainer interop.Hash160, keys []interop.PublicKey) {
 	ctx := storage.GetContext()
 
 	if !common.HasUpdateAccess(ctx) {
 		panic("only owner can reinitialize contract")
+	}
+
+	if len(addrBalance) != 20 || len(addrContainer) != 20 {
+		panic("init: incorrect length of contract script hash")
 	}
 
 	var irList []common.IRNode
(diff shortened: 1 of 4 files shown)


netmap/netmap_contract.go
```diff
@@ -53,13 +57,21 @@ var (
 
 // Init function sets up initial list of inner ring public keys and should
 // be invoked once at neofs infrastructure setup.
-func Init(owner interop.Hash160, keys []interop.PublicKey) {
+func Init(
+	owner interop.Hash160,
+	addrBalance interop.Hash160,
+	addrContainer interop.Hash160,
+	keys []interop.PublicKey) {
```
cthulhu-rider, line 64: Can we
```
func Init(owner, addrBalance, addrContainer interop.Hash160, keys []interop.PublicKey)
```
?

cthulhu-rider (approved):

netmap/netmap_contract.go
alexvanin: Sure, thx. Rushed on this.

PR #63, opened 2021-03-25
Fix threshold number calculation

alexvanin:
`x*2/3` is not equal to `x/3*2` with integers. The only correct way to calculate threshold is the first one.


cthulhu-rider (approved):
Rather repetitive code, perhaps in the future it is worth moving it into a separate function

alexvanin:
> Rather repetitive code, perhaps in the future it is worth moving it into a separate function

Somehow it is hidden in `ir.go` and every side chain contract re-use it. Hope one day NeoFS contract will catch it up.

PR #64, opened 2021-03-29
Fix init arguments in contracts

alexvanin:
With changes in governance and multi signature collection via notary contract, several init arguments become redundant.


alexvanin:
According to decisions on notary support, this PR is not valid anymore.

PR #68, opened 2021-04-07
Unify transfer deatails format

alexvanin:
Unified format uses transfer type as the first byte and extra details next. List of transfer types used in contracts defined in `transfer.go`. It includes:
- mint,
- burn,
- lock,
- unlock,
- container fee.

### DIFF
--- a/container/container_contract.go
+++ b/container/container_contract.go
@@ -50,8 +50,7 @@ const (
 )
 
 var (
-	containerFeeTransferMsg = []byte("container creation fee")
-	eACLPrefix              = []byte("eACL")
+	eACLPrefix = []byte("eACL")
 )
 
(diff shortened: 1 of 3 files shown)


container/container_contract.go
```diff
@@ -122,13 +122,14 @@ func Put(container []byte, signature interop.Signature, publicKey interop.Public
 	for i := 0; i < len(alphabet); i++ {
 		node := alphabet[i]
 		to := contract.CreateStandardAccount(node.PublicKey)
+		details := common.ContainerFeeTransferDetails(containerID)
```
cthulhu-rider, line 125: Can we do it before loop once?

cthulhu-rider (approved):

container/container_contract.go
alexvanin: Yes, good catch.

PR #74, opened 2021-04-20
Support notary and w/o notary work flows in contracts

### LINKED ISSUE #70: Transfer assets to alphabet nodes at neofs.Withdraw
To withdraw assets, alphabet nodes should accept it by invoking neofs.Cheque method in main net. This costs real GAS and should be payed by someone. We can take fee from the withdraw invoker. 

To do:
- [ ] find out max cost of cheque operation (see #69 to reduce cost),
- [ ] transfer this cost to each alphabet node at neofs.Withdaw.

alexvanin:
In this pull request:
- contracts supported enabled and disabled notary work flows: if notary enabled at init stage, then contracts use multi signature checks, otherwise contracts collect ballots,
- added `processing` contract that processes multi signature checks in main chain and pays for them when notary enabled,
- added withdraw fee, that transferred to `processing` contract or alphabet nodes to pay for `neofs.Cheque` invocation,
- all default global configuration records are gone, they should be specified by initialization scripts external to contracts.

Using multi signature checks allows to decrease withdraw fee. With ballot collection it required from `0.4*7=2.8` up to `7.0` GAS per withdraw. With multi signature checks it is about `0.2` GAS per withdraw. 

Closes #70


alexvanin:
Workflow with notary before:
image

Workflow with optional notary after:
image

PR #80, opened 2021-05-17
neofs: remove `deposit` method

### LINKED ISSUE #71: Drop 'deposit' method
1) It's not really needed, all the real logic is implemented in `OnNEP17Payment`, `maxBalanceAmount` check could just be moved there.
2) `data` is "" most of the time there anyway, so most people won't notice, but we will eventually provide a way to specify that too (nspcc-dev/neo-go#1896).
3) It's problematic wrt witness scoping, for it to work one either specifies "Global" scope or contract-specific scope, the default "CalledByEntry" doesn't work for it.

fyrchik:
Close #71 .
Testing is in process.

### DIFF
--- a/neofs/neofs_contract.go
+++ b/neofs/neofs_contract.go
@@ -8,7 +8,6 @@ package smart_contract
 	- InitConfig
 
 	User related methods:
-	- Deposit
 	- Withdraw
 	- Bind
 	- Unbind
@@ -67,6 +66,7 @@ const (
 	publicKeySize = 33
 
 	maxBalanceAmount = 9000 // Max integer of Fixed12 in JSON bound (2**53-1)
+	maxBalanceAmountGAS = maxBalanceAmount * 1_0000_0000
 
 	// hardcoded value to ignore deposit notification in onReceive
 	ignoreDepositNotification = "\x57\x0b"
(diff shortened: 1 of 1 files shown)


neofs/config.yml
```diff
@@ -1,16 +1,6 @@
 name: "NeoFS"
 safemethods: ["alphabetList", "alphabetAddress", "innerRingCandidates", "config", "listConfig", "version"]
 events:
-  - name: Deposit
```
alexvanin, line 4: This event produced by `OnNEP17Payment` function to track deposits, thus it is still there.

alexvanin (changes requested):

neofs/neofs_contract.go
```diff
@@ -253,6 +252,12 @@ func OnNEP17Payment(from interop.Hash160, amount int, data interface{}) {
 		return
 	}
 
+	if amount <= 0 {
+		panic("onNEP17Payment: amount must be positive")
+	} else if maxBalanceAmount < amount {
```
alexvanin, line 257: Update `maxBalanceAmount` constant. It should be 10^8 times bigger value.

alexvanin (changes requested):

neofs/neofs_contract.go
fyrchik: Fixed.

alexvanin (approved):

PR #83, opened 2021-05-21
[#79] Panic instead of returning bool value

### LINKED ISSUE #79: Raise an exception instead of returning `bool` value
Currently many methods return `bool` value as a signal of success. In most of the cases just `true` value is returned. This is unneeded and we can raise an exception instead. This will also simplify response handling on caller side (either client or other contract). For example, `netmap` contract doesn't even check return values of `newEpoch` invocations of `balance` and `container` contracts. We can `panic` everywhere for consistency.

fyrchik:
There is a number of contracts which return only `true` value.
Also handling `FAULT` on the client is easier then also checking return
value.

Close #79 .

### DIFF
--- a/balance/balance_contract.go
+++ b/balance/balance_contract.go
@@ -126,7 +126,7 @@ func Transfer(from, to interop.Hash160, amount int, data interface{}) bool {
 	return token.transfer(ctx, from, to, amount, false, nil)
 }
 
-func TransferX(from, to interop.Hash160, amount int, details []byte) bool {
+func TransferX(from, to interop.Hash160, amount int, details []byte) {
 	ctx := storage.GetContext()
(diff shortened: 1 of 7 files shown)


balance/balance_contract.go
```diff
@@ -272,8 +267,6 @@ func NewEpoch(epochNum int) bool {
 			token.transfer(ctx, addr, acc.Parent, acc.Balance, true, details)
 		}
 	}
-
-	return true
 }
 
 func Mint(to interop.Hash160, amount int, txDetails []byte) bool {
```
alexvanin, line 272: `Mint` and `Burn` methods are not the part of NEP-17. Maybe it makes sense to remove return type there as well?

fyrchik: Yes, wiil do this.

alexvanin (approved):

PR #109, opened 2021-07-21
Fix/Do not emit GAS to `proxy` if notary disabled

### LINKED ISSUE #106: Produce more GAS to alphabet nodes in non notary deploys
If notary service is not enabled in side chain, then we should produce more gas to alphabet nodes.

carpawell:
Closes #106.


alexvanin:
Commit headers should have `#` sign in brackets.

alexvanin (approved):

PR #127, opened 2021-09-20
Remove contract owners

### LINKED ISSUE #107: Allow contract migration to multisigned transaction from Alphabet
`Migrate` methods can be invoked only by "contract owner", which is set manually at deploy. This is not quite secure to have single wallet with contract owner, because this wallet can be unrelated to NeoFS in the first place. Instead we can check witness of alphabet multi signature before contract migration. The other option is to check committee witness. 

In this case we can remove `owner` argument from contract deployment routines.

alexvanin:
Do not save contract owner at the contract deploy stage. Use side chain committee witness to migrate contracts.
Close #107

### DIFF
--- a/common/ir.go
+++ b/common/ir.go
@@ -55,6 +55,12 @@ func AlphabetAddress() []byte {
 	return Multiaddress(alphabet, false)
 }
 
+// CommitteeAddress returns multi address of committee.
+func CommitteeAddress() []byte {
+	committee := neo.GetCommittee()
+	return Multiaddress(committee, true)
+}
+
 // Multiaddress returns default multi signature account address for N keys.
 // If committee set to true, then it is `M = N/2+1` committee account.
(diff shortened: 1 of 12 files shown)


carpawell:
@alexvanin, conflicts

common/ir.go
```diff
@@ -63,12 +69,7 @@ func Multiaddress(n []interop.PublicKey, committee bool) []byte {
 		threshold = len(n)/2 + 1
 	}
 
-	keys := []interop.PublicKey{}
```
carpawell, line 66: why did we have such a transformation? and why it becomes useless now?

alexvanin: As far as I remember, `neo.GetCommittee()` returned `[][]byte` instead of `[]interop.PublicKey{}` so we had to do these, but I am not 100% sure.

carpawell (approved):

PR #133, opened 2021-09-28
[#125] Zone transfer

### LINKED ISSUE #125: Support zone transfer in NNS contract
To implement zone transfer in [coredns](https://github.com/nspcc-dev/coredns) we need `nns contract` support it too.

At least we need:
* Appropriate [SOA records](https://en.wikipedia.org/wiki/SOA_record)

KirillovDenis:
Added:
* Support multiple records of the same type
* SOA record

closes #125 

Signed-off-by: Denis Kirillov <denis@nspcc.ru>

### DIFF
--- a/nns/nns_contract.go
+++ b/nns/nns_contract.go
@@ -62,8 +62,8 @@ const (
 const (
 	// defaultRegisterPrice is the default price for new domain registration.
 	defaultRegisterPrice = 10_0000_0000
-	// millisecondsInYear is amount of milliseconds per year.
-	millisecondsInYear = 365 * 24 * 3600 * 1000
+	// millisecondsInSecond is amount of milliseconds per second.
+	millisecondsInSecond = 1000
 )
 
 // RecordState is a type that registered entities are saved to.
@@ -242,7 +242,7 @@ func IsAvailable(name string) bool {
 }
 
 // Register registers new domain with the specified owner and name if it's available.
-func Register(name string, owner interop.Hash160) bool {
+func Register(name string, owner interop.Hash160, email string, refresh, retry, expire, ttl int) bool {
 	fragments := splitAndCheck(name, false)
 	if fragments == nil {
 		panic("invalid domain name format")
@@ -277,23 +277,24 @@ func Register(name string, owner interop.Hash160) bool {
 	ns := NameState{
 		Owner:      owner,
 		Name:       name,
-		Expiration: runtime.GetTime() + millisecondsInYear,
+		Expiration: runtime.GetTime() + expire*millisecondsInSecond,
 	}
 	putNameStateWithKey(ctx, tokenKey, ns)
+	putSoaRecord(ctx, name, email, refresh, retry, expire, ttl)
 	updateBalance(ctx, []byte(name), owner, +1)
 	postTransfer(oldOwner, owner, []byte(name), nil)
 	return true
 }
 
 // Renew increases domain expiration date.
-func Renew(name string) int {
+func Renew(name string, expire int) int {
 	if len(name) > maxDomainNameLength {
 		panic("invalid domain name format")
 	}
 	runtime.BurnGas(GetPrice())
 	ctx := storage.GetContext()
 	ns := getNameState(ctx, []byte(name))
-	ns.Expiration += millisecondsInYear
+	ns.Expiration += expire * millisecondsInSecond
 	putNameState(ctx, ns)
 	return ns.Expiration
 }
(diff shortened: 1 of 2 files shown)


nns/nns_contract.go
```diff
@@ -468,20 +472,29 @@ func putNameStateWithKey(ctx storage.Context, tokenKey []byte, ns NameState) {
...
+
+	var result []string
+	records := storage.Find(ctx, recordsKey, storage.DeserializeValues)
+	for iterator.Next(records) {
+		r := iterator.Value(records).(struct {
+			key string
+			rs  RecordState
+		})
+		value := r.rs.Data
+		rTyp := r.key[len(r.key)-5]
```
fyrchik, line 488: Why not `r.rs.Type`?

nns/nns_contract.go
```diff
@@ -468,20 +472,29 @@ func putNameStateWithKey(ctx storage.Context, tokenKey []byte, ns NameState) {
...
 	}
-	record := std.Deserialize(recBytes.([]byte)).(RecordState)
-	return record.Data
+	return result
 }
 
 // putRecord stores domain record.
 func putRecord(ctx storage.Context, tokenId []byte, name string, typ RecordType, record string) {
-	recordKey := getRecordKey(tokenId, name, typ)
+	recordKey := addUniqueBytesToRecordKey(getRecordKey(tokenId, name, typ))
```
fyrchik, line 498: How do I replace an existing record with this API? I expect `SetRecord` (e.g. of both A and CNAME type) to replace old record, not to add new IP along with the old (possibly invalid one). I think `AddRecord` can use this logic and `SetRecord` (`SetRecords` to replace a set of records).

nns/nns_contract.go
```diff
@@ -468,20 +472,29 @@ func putNameStateWithKey(ctx storage.Context, tokenKey []byte, ns NameState) {
...
 }
 
-// getRecord returns domain record.
-func getRecord(ctx storage.Context, tokenId []byte, name string, typ RecordType) string {
-	recordKey := getRecordKey(tokenId, name, typ)
-	recBytes := storage.Get(ctx, recordKey)
-	if recBytes == nil {
-		return recBytes.(string) // A hack to actually return NULL.
+// getRecords returns domain record.
+func getRecords(ctx storage.Context, tokenId []byte, name string, typ RecordType) []string {
```
fyrchik, line 477: `getRecordsByType`?

nns/nns_contract.go
```diff
@@ -503,6 +516,13 @@ func getRecordKey(tokenId []byte, name string, typ RecordType) []byte {
 	return append(recordKey, []byte{byte(typ)}...)
 }
 
+// addUniqueBytesToRecordKey add to key 4 unique bytes to allow store several records with the same type.
+func addUniqueBytesToRecordKey(recordKey []byte) []byte {
+	tx := runtime.GetScriptContainer()
+	bb := []byte(tx.Hash)
+	return append(recordKey, bb[:4]...)
+}
```
fyrchik, line 544: I think, maintaining a counter is more reliable. This implementation also prohibits us from setting multiple records in a single transaction (which is unexpected behaviour from user POV).

fyrchik: If we stay with the random numbers, we better use `runtime.GetRandom`, as it solves the problem with multiple queries in a single tx.

nns/recordtype.go
```diff
@@ -9,6 +9,8 @@ const (
 	A RecordType = 1
 	// CNAME represents canonical name record type.
 	CNAME RecordType = 5
+	// SOA represents start of a zone of authority record type.
```
fyrchik, line 12: /a zone of/d

alexvanin:
I suppose we need to add migration routine for such change :thinking: 
Or it should work just fine on unique records without `serial`?

KirillovDenis:
> I suppose we need to add migration routine for such change thinking Or it should work just fine on unique records without `serial`?

The `RecordState` struct was changed so It seems we have to add some migration.

nns/nns_contract.go
```diff
@@ -68,9 +68,10 @@ const (
 
 // RecordState is a type that registered entities are saved to.
 type RecordState struct {
-	Name string
-	Type RecordType
-	Data string
+	Name   string
+	Type   RecordType
+	Data   string
+	Serial byte
```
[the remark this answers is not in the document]
alexvanin: https://github.com/nspcc-dev/neofs-contract/pull/133#discussion_r717450842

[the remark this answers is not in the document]
KirillovDenis: > No need to set serial. We can use block height and transaction number/id instead.

But what we should do if we want to change specific record? How we can know what exactly block number is contained in the record key?

KirillovDenis:
> I suppose we need to add migration routine for such change thinking Or it should work just fine on unique records without `serial`?

Do we need to generate default SOA records during migration? Or domain owner will add it later using the `updateSOA` method?

alexvanin:
> > I suppose we need to add migration routine for such change thinking Or it should work just fine on unique records without `serial`?
> 
> Do we need to generate default SOA records during migration? Or domain owner will add it later using the `updateSOA` method?

I think yes. In fact, they can be set with NSPCC related values. 
@realloc What SOA records do we want for `<contract>.neofs` domains?

nns/nns_contract.go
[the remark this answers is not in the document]
realloc: Serial is for the zone, not for the record. It is updated with every record change in the zone.

realloc:
> What SOA records do we want for `<contract>.neofs` domains?

name - zone name
email - ops@nspcc.ru
refresh - 86400
retry - 7200
expire - ∞
ttl - 3600

alexvanin (approved):

PR #135, opened 2021-09-29
Register nice names for new containers

### LINKED ISSUE #134: Register container nice names in NNS contract
Users want to access containers outside of the core NeoFS by nice names. Unique bucket names in S3 are good example of nice names we are lack of. However, with NNS contract it is possible to register container names as the domains in separate zones. Then third party software as S3 Gate or SDK client can resolve it via NNS or even DNS requests to a NNS backed coredns server.

## NeoFS API

NeoFS API should define two well-known container attributes:
- `__NEOFS__NAME`
- `__NEOFS__ZONE`

These attributes may be set up by third party apps or libraries. They register nice name in one of the specified zones. Zone examples:
- .neofs (restricted)
- .cdn
- .s3

## Container contract

At first, container contract should be provided with NNS contract address at deploy stage.

Then container contract can be expanded with `PutWithName()` function:
```go
PutWithName(container []byte, signature interop.Signature, publicKey interop.PublicKey, token []byte, name, zone string)
```

Storage node should invoke this function if container has `__NEOFS__NAME` and `__NEOFS__ZONE` attributes.

Before container creation, function can do small sanity check:
- if zone available
- if name+zone is available

Then create a container as it was done before. Contract can charge +10% more Alphabet fee for nice name registration. In the end invoke register method on `name+zone` domain and store container ID as TXT record there. The domain owner can be the container contract itself.

At delete operation, container contract removes TXT record. 

## Possible issues
- Is it possible to use container contract as the owner of these domains?
- How to renew the domains? Should we?
- Is it possible for random users do the cybersquatting in NNS contract at side chain (GAS is a bit worthless there) and if so, how we attend that.

fyrchik:
Close #134 .

Adressing some issues:
1. Container contract is the owner of all domains.
2. Renewing can be done via direct invocation of NNS contract. Currently there is no owner check, so anyone (alphabet) can pay GAS to renew domain.
3. Cybersquatting is possible, because anyone can register a domain. One way to prevent this is to allow registering "restricted" roots which need a committee signature to register domains in.

### DIFF
--- a/container/container_contract.go
+++ b/container/container_contract.go
@@ -46,6 +46,8 @@ const (
 	neofsIDContractKey = "identityScriptHash"
 	balanceContractKey = "balanceScriptHash"
 	netmapContractKey  = "netmapScriptHash"
+	nnsContractKey     = "nnsScriptHash"
+	nnsRootKey         = "nnsRoot"
 	notaryDisabledKey  = "notary"
 
 	containerFeeKey = "ContainerFee"
@@ -61,6 +63,10 @@ var (
 	eACLPrefix = []byte("eACL")
 )
 
+// OnNEP11Payment is needed for registration with contract as owner to work.
+func OnNEP11Payment(a interop.Hash160, b int, c []byte, d interface{}) {
+}
+
 func _deploy(data interface{}, isUpdate bool) {
 	ctx := storage.GetContext()
 
@@ -73,6 +79,8 @@ func _deploy(data interface{}, isUpdate bool) {
 	addrNetmap := args[1].(interop.Hash160)
 	addrBalance := args[2].(interop.Hash160)
 	addrID := args[3].(interop.Hash160)
+	addrNNS := args[4].(interop.Hash160)
+	nnsRoot := args[5].(string)
 
 	if len(addrNetmap) != 20 || len(addrBalance) != 20 || len(addrID) != 20 {
 		panic("incorrect length of contract script hash")
@@ -81,6 +89,8 @@ func _deploy(data interface{}, isUpdate bool) {
 	storage.Put(ctx, netmapContractKey, addrNetmap)
 	storage.Put(ctx, balanceContractKey, addrBalance)
 	storage.Put(ctx, neofsIDContractKey, addrID)
+	storage.Put(ctx, nnsContractKey, addrNNS)
+	storage.Put(ctx, nnsRootKey, nnsRoot)
 
 	// initialize the way to collect signatures
 	storage.Put(ctx, notaryDisabledKey, notaryDisabled)
@@ -88,6 +98,9 @@ func _deploy(data interface{}, isUpdate bool) {
 		common.InitVote(ctx)
 		runtime.Log("container contract notary disabled")
 	}
+
+	// add NNS root for container alias domains
+	contract.Call(addrNNS, "addRoot", contract.All, nnsRoot)
 
 	runtime.Log("container contract initialized")
 }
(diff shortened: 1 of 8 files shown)


container/container_contract.go
```diff
@@ -182,6 +205,31 @@ func Put(container []byte, signature interop.Signature, publicKey interop.Public
...
+			res := contract.Call(nnsContractAddr, "register", contract.All, domain, runtime.GetExecutingScriptHash()).(bool)
+			if !res {
+				panic("can't register the domain " + domain)
+			}
+		} else {
+			res := contract.Call(nnsContractAddr, "getRecord", contract.ReadStates|contract.AllowCall, domain, 16 /* TXT */)
+			if res != nil {
+				panic("name is already taken")
+			}
+		}
```
alexvanin, line 225: Notary disabled environment will benefit if sanity check will happen before notification. In this case, test invocation will fail before Storage node will return response to the client. Therefore client will get nice error "container name is already taken").

For example we have balance check up there for this reason:
```go
	if !alphabetCall {
		if balance < containerFee*len(alphabet) {
			panic("insufficient balance to create container")
		}
		runtime.Notify("containerPut", container, signature, publicKey, token)
		return
	}
```

container/container_contract.go
```diff
@@ -112,6 +127,14 @@ func Update(script []byte, manifest []byte, data interface{}) {
 // Token is optional and should be stable marshaled SessionToken structure from
 // API.
 func Put(container []byte, signature interop.Signature, publicKey interop.PublicKey, token []byte) {
+	PutNamed(container, signature, publicKey, token, "", "")
+}
+
+// PutNamed is similar to put but also sets a CNAME record in nns contract.
```
alexvanin, line 133: > also sets a CNAME record

The code sets TXT record.

alexvanin:
> One way to prevent this is to allow registering "restricted" roots which need a committee signature to register domains in.

Notary disabled environment can't form committee signature :cry:

container/container_contract.go
```diff
@@ -237,6 +285,20 @@ func Delete(containerID []byte, signature interop.Signature, token []byte) {
 		common.RemoveVotes(ctx, id)
 	}
 
+	key := append([]byte(nnsHasAliasKey), containerID...)
+	domain := storage.Get(ctx, key).(string)
+	if len(domain) != 0 {
+		storage.Delete(ctx, key)
+		// We should do `getRecord` first because NNS record could be deleted
+		// by other means (expiration, manual) thus leading to failing `deleteRecord`
```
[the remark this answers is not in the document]
fyrchik: Yes, implicitly (it takes struct and checks owner, which should panic for null).

fyrchik:
Integer literals are used in imported constants because of a compiler warning. We should fix this after https://github.com/nspcc-dev/neo-go/pull/2205 .

container/container_contract.go
fyrchik: What's the reason for this behaviour? I thought failed execution should have no effect on subscribers.

alexvanin: > I thought failed transactions should have no effect on subscribers.

Indeed, they have no effect on subscribers. The reason for this behavior -- _at notary disabled environment_ return error to the client as soon as possible for better UX. 

Compare these two cases:

1. Domain name sanity check at container creation:
image


2. Domain name sanity check before `containerPut` notification.
image

fyrchik:
Fixed comments, let's not merge I will write some tests.

container/container_contract.go
```diff
@@ -182,13 +219,46 @@ func Put(container []byte, signature interop.Signature, publicKey interop.Public
 
 	addContainer(ctx, containerID, ownerID, cnr)
 
+	if name != "" {
+		if needRegister {
```
alexvanin, line 223: `if name != "" && needRegister` will be easier to read with one less indent, probably

alexvanin (approved):
Seems good. `PutNamed` comment still mentions `CNAME` instead of `TXT`, though.

container/container_contract.go
fyrchik: There is another logic after the inner `if`, I would prefer to leave it as it is.

PR #137, opened 2021-10-02
Implement contract testing framework

fyrchik:
It supports everything we need, including cross-contract calls. The only exception is testing contract update, because it requires having 2 different versions of contract. We may support this in future after some discussion.


fyrchik:
I think CDN contract may have similar tests. We may export our helpers to make things easy.

alexvanin:
> I think CDN contract may have similar tests. We may export our helpers to make things easy.

To do that we need to make `util_test.go` functions public, I guess.

alexvanin:
Add `make test` target to run tests and mention it in README file.

alexvanin (approved):

alexvanin (approved):

PR #141, opened 2021-10-07
Fixes for contract update

fyrchik:
Container contract must take into account that NNS root can already be registered (we will need this code for the following updates anyway).


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/nspcc-dev/neofs-contract
 go 1.14
 
 require (
+	github.com/mr-tron/base58 v1.2.0
```
alexvanin, line 6: Accidentally?

alexvanin: Missed that in `container_test.go`

alexvanin (approved):

PR #144, opened 2021-10-12
reputation: fix safe method names

fyrchik:
Also make it possible to compile with the latest neo-go version.


alexvanin (approved):
Wonder how it even worked before.

Issue #101, opened 2021-06-30
Make simultanious put methods stable


alexvanin:
If contract method update some slice of structures in storage, then we should check if it is possible to have multiple transaction in one block changing the same slice. If so, then we should store list as a k:v pairs with some common prefix. See example in  https://github.com/nspcc-dev/neofs-contract/pull/100

## Reputation contract

- `reputation.Put`: multiple reputation of one peer will fail in a single block.
```go
	reputationValues := GetByID(id)
	reputationValues = append(reputationValues, value)

	rawValues := std.Serialize(reputationValues)

	storage.Put(ctx, id, rawValues)
```

## Container contract

- `container.PutCotnainerSize`: multiple estimations of one container will fail in a single block.
```gp
	key := estimationKey(epoch, cid)
	s := getContainerSizeEstimation(ctx, key, cid)

	s.estimations = append(s.estimations, estimation{
		from: pubKey,
		size: usedSize,
	})

	storage.Put(ctx, key, std.Serialize(s))
```

- `container.Put`: multiple containers from new owners will fail in a single block.
```go
	addOrAppend(ctx, ownersKey, owner)
``` 

- `container.Put`: multiple containers from one owner will fail in a single block.
```go
	addOrAppend(ctx, owner, id)
```

The same applies for `container.Delete`.

## NeoFS contract

- `neofs.InnerRingCandidateAdd`, `neofs.InnerRingCandidateRemove`: multiple invocations from different candidates will fail in a single block.
```go
	candidates := getNodes(ctx, candidatesKey)

	list, ok := addNode(candidates, c)
	if !ok {
		panic("irCandidateAdd: candidate already in the list")
	}

	common.SetSerialized(ctx, candidatesKey, list)
```

## NeoFSID contract
- `neofsid.AddKey`, `neofsid.RemoveKey`: multiple inovcation for the same owner will fail in a single block.
```go
	info := getUserInfo(ctx, owner)

addLoop:
	for i := 0; i < len(keys); i++ {
		pubKey := keys[i]
		info.Keys = append(info.Keys, pubKey)
	}

	common.SetSerialized(ctx, owner, info)
```

--- 

Some of these cases are quite critical and others are not. I propose to fix container contract issues for v0.95.2 release and then fix others later by this order:

0. Container contract
1. Reputation contract
2. NeoFSID contract
3. NeoFS contract

cthulhu-rider:
Needed to be revised.

[...]

PR #166, opened 2021-11-16
neofs: allow to register multiple candidates per block

fyrchik:
Related #101.

### DIFF
--- a/neofs/neofs_contract.go
+++ b/neofs/neofs_contract.go
@@ -23,7 +23,8 @@ type (
 )
 
 const (
-	candidateFeeConfigKey = "InnerRingCandidateFee"
+	// CandidateFeeConfigKey contains fee for a candidate registration.
+	CandidateFeeConfigKey = "InnerRingCandidateFee"
 	withdrawFeeConfigKey  = "WithdrawFee"
 
 	alphabetKey       = "alphabet"
@@ -48,6 +49,14 @@ var (
 // _deploy sets up initial alphabet node keys.
 func _deploy(data interface{}, isUpdate bool) {
 	if isUpdate {
+		ctx := storage.GetContext()
+		nodes := getNodes(ctx, candidatesKey)
+		storage.Delete(ctx, candidatesKey)
+
+		for i := range nodes {
+			key := append([]byte(candidatesKey), nodes[i].PublicKey...)
+			storage.Put(ctx, key, []byte{1})
+		}
 		return
 	}
 
@@ -79,7 +88,6 @@ func _deploy(data interface{}, isUpdate bool) {
 
 	// initialize all storage slices
 	common.SetSerialized(ctx, alphabetKey, irList)
-	common.SetSerialized(ctx, candidatesKey, []common.IRNode{})
 
 	storage.Put(ctx, processingContractKey, addrProc)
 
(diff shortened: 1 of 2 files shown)


neofs/neofs_contract.go
```diff
@@ -212,24 +220,26 @@ func InnerRingCandidateAdd(key interop.PublicKey) {
 	}
 
 	c := common.IRNode{PublicKey: key}
-	candidates := getNodes(ctx, candidatesKey)
-
-	list, ok := addNode(candidates, c)
```
alexvanin, line 217: `addNode` function can be removed.

neofs/neofs_contract.go
```diff
@@ -568,6 +578,17 @@ func setConfig(ctx storage.Context, key, val interface{}) {
 	storage.Put(ctx, storageKey, val)
 }
 
+func hasNode(ctx storage.Context, key []byte, n common.IRNode) bool {
+	rawNode := std.Serialize(n)
+	it := storage.Find(ctx, key, storage.ValuesOnly)
+	for iterator.Next(it) {
+		if common.BytesEqual(rawNode, iterator.Value(it).([]byte)) {
+			return true
```
alexvanin, line 586: Why `hasNode` compares serialized `common.IRNode{}` structures? It seems that now contract stores only public keys of those nodes. I suppose it should be 

```go
prefix := []byte(candidatesKey)
stKey := append(prefix, n.PublicKey...)
return storage.Get(ctx, stKey) != nil
```

neofs/neofs_contract.go
```diff
@@ -212,24 +220,26 @@ func InnerRingCandidateAdd(key interop.PublicKey) {
...
 
 	transferred := gas.Transfer(from, to, fee, []byte(ignoreDepositNotification))
 	if !transferred {
 		panic("irCandidateAdd: failed to transfer funds, aborting")
 	}
 
+	stKey := append([]byte(candidatesKey), key...)
+	if storage.Get(ctx, stKey) != nil {
+		panic("irCandidateAdd: candidate already in the list")
+	}
```
alexvanin, line 239: Considering comment about `hasNode`, I am a bit confused. Do we need to check it there if `hasNode` should do it instead? Maybe we do not need `hasNode` at all with this check?

neofs/neofs_contract.go
fyrchik: Fixed. My initial implementation was governed by the fact that `InnerRingCandidateAdd` can, in theory, update node info if the `IRNode` struct will be extended in future. Though old one compared public keys only so we are ok I thing.

neofs/neofs_contract.go
fyrchik: Removed.

neofs/neofs_contract.go
```diff
@@ -172,16 +187,11 @@ func InnerRingCandidateRemove(key interop.PublicKey) {
...
-		if !common.BytesEqual(c.PublicKey, key) {
-			nodes = append(nodes, c)
-		} else {
-			runtime.Log("irCandidateRemove: candidate has been removed")
-		}
+	prefix := []byte(candidatesKey)
+	stKey := append(prefix, key...)
+	if storage.Get(ctx, stKey) != nil {
+		storage.Delete(ctx, stKey)
+		runtime.Log("irCandidateRemove: candidate has been removed")
```
alexvanin, line 194: This code should be executed after `common.RemoveVotes(ctx, hashID)`

For notary disabled environment, there is a pattern to prepare data before (to decrease extra GAS addition) `common.RemoveVotes(ctx, hashID)` but store in in the contract memory after. This is way we were preparing new list there, but actually update it (`common.SetSerialized()`) in the end.

neofs/neofs_contract.go
alexvanin: Still there!

neofs/neofs_contract.go
alexvanin: `hasNode` has no references now, remove?

alexvanin (changes requested):

neofs/neofs_contract.go
fyrchik: Fixed.

alexvanin (approved):

alexvanin:
Rebase, please.

PR #168, opened 2021-11-17
netmap: fix tests

fyrchik:


alexvanin (approved):
Part of #161? Would love to see it in commit header.

PR #170, opened 2021-11-17
nns: disallow record duplicates

### LINKED ISSUE #165: Should NNS contract control unique record values?
NNS contract does not check if the record has unique value.

```
$ neo-go contract invokefunction -r [endpoint] -w [wallet] -a [address] [nns] addRecord string:neofs.org int:1 string:1.1.1.1 -- [witness]
Sent invocation transaction 9d6afe41af27110b7c89accb0f8ab1523d218015a32a68fd223cc5689d7f13b6

$ neo-go contract invokefunction -r [endpoint] -w [wallet] -a [address] [nns] addRecord string:neofs.org int:1 string:1.1.1.1 -- [witness]
Sent invocation transaction c24cab40869e23d079288f9c8599b4f1d6da3ea8ca3b4cc16d1f0870ece611e4

$ neo-go contract testinvokefunction -r [endpoint] [nns] getRecords string:neofs.org int:1
{
  "state": "HALT",
  "gasconsumed": "30651030",
  "script": "EQwJbmVvZnMub3JnEsAfDApnZXRSZWNvcmRzDBS0b+igJfsitF5g10KILOetqciQ1kFifVtS",
  "stack": [
    {
      "type": "Array",
      "value": [
        {
          "type": "ByteString",
          "value": "MS4xLjEuMQ=="
        },
        {
          "type": "ByteString",
          "value": "MS4xLjEuMQ=="
        }
      ]
    }
  ]
}
```

However I can deploy coreDNS server that serves two same `A` records as well. 

```
$ cat example.org 
$ORIGIN example.org.
@       3600 IN SOA sns.dns.icann.org. noc.dns.icann.org. 2017042745 7200 3600 1209600 3600
        3600 IN NS a.iana-servers.net.
        3600 IN NS b.iana-servers.net.

www     IN A     127.0.0.1
        IN A     127.0.0.1
        IN AAAA  ::1

$ ./coredns -conf cfg &
[1] 1515459

$ dig @localhost -p 5300 a www.example.org
www.example.org.        3600    IN      A       127.0.0.1
www.example.org.        3600    IN      A       127.0.0.1
```

It doesn't seem right to me, but it looks like a valid behavior. Should the caller perform sanity checks to avoid duplicates?

fyrchik:
Close #165 .
So we already iterate over records to checking for duplicates doesn't seem to introduce a big overhead.

### DIFF
--- a/nns/nns_contract.go
+++ b/nns/nns_contract.go
@@ -584,9 +584,14 @@ func addRecord(ctx storage.Context, tokenId []byte, name string, typ RecordType,
 	recordsKey := getRecordsKeyByType(tokenId, name, typ)
 
 	var id byte
-	records := storage.Find(ctx, recordsKey, storage.KeysOnly)
+	records := storage.Find(ctx, recordsKey, storage.ValuesOnly|storage.DeserializeValues)
 	for iterator.Next(records) {
 		id++
+
+		r := iterator.Value(records).(RecordState)
+		if r.Name == name && r.Type == typ && r.Data == data {
+			panic("record already exists")
+		}
 	}
 
 	if typ == CNAME && id != 0 {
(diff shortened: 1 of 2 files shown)


alexvanin (approved):
Can you rebase once again after https://github.com/nspcc-dev/neofs-contract/pull/168

Issue #122, opened 2021-09-09
Add subnetwork contract


alexvanin:
## Summary
Create new `Subnet` contract for NeoFS sidechain which provides the functionality to create and work with subnets. It works with subnet IDs and info structures defined in NeoFS API V2 protocol (https://github.com/nspcc-dev/neofs-api/issues/180). Also operates with public keys and owner IDs.

## Task

Create `Subnet` contract with methods;

- [x] `Put(bytes id, bytes owner, bytes info)` method. It must check uniqueness of the identifier. If called by required number of alphabet nodes, saves the subnet, otherwise produces corresponding notification.

- [x] `Delete(bytes ID)` method. It must require the method to be called by the subnet owner only. It can fail if subnet is missing.

- [x] `AddNodeAdmin(bytes subnetID, interop.PublicKey adminPublicKey)` which extends the list of admin keys which are allowed to manage whitelist of subnet nodes. It must require the method to be called by subnet owner only.

- [x] `RemoveNodeAdmin(bytes subnetID, interop.PublicKey adminPublicKey)` which is same as `AddNodeAdmin` but removal.

- [x] `AddNode(bytes subnetID, interop.PublicKey publicKey)` which extends the list of node IDs which are allowed to entry the subnet. It must require the method to be called by subnet owner or node admin only.

- [x] `RemoveNode(bytes subnetID, interop.PublicKey publicKey)` which is same as `AddNode` but with removal. Produces corresponding notification. Produces corresponding notification.

- [x] `ListNodes(bytes subnetID) []bytes` which returns list of IDs added via `AddNode`.

- [x] `AddClientAdmin(bytes subnetID, bytes groupID, interop.PublicKey adminPublicKey)` which appends the list of admin keys for the specific group which are allowed to manage whitelist of subnet clients. It must require the method to be called by subnet owner only.

- [x] `RemoveClientAdmin(bytes subnetID, bytes groupID, interop.PublicKey adminPublicKey)` which is same as `AddClientAdmin` but removal.

- [x] `AddUser(bytes subnetID, bytes groupID, bytes ownerID) ` which extends list of users which are allowed to bind their containers to the subnet. It must require the method to be called by subnet owner or client group admin only.

- [x]  `RemoveUser(bytes subnetID, bytes groupID, bytes ownerID)` which is same as `AddUser` but with removal.

- [x] `ListUsers(bytes subnetID) []bytes` which returns list of IDs added via `AddUser`.

After that:

[...]

PR #171, opened 2021-11-18
Initial subnet contract implementation

fyrchik:
Support put/get/delete methods.

Related #122 .
Depends on #168 .

### DIFF
--- a/subnet/config.yml
+++ b/subnet/config.yml
@@ -2,3 +2,11 @@ name: "NeoFS Subnet"
 safemethods: ["version"]
 permissions:
   - methods: ["update"]
+events:
+  - name: SubnetPut
+    parameters:
+      - name: ownerKey
+        type: PublicKey
+      - name: info
+        type: ByteArray
+
--- a/subnet/config.yml
+++ b/subnet/config.yml
@@ -1,0 +1,4 @@
+name: "NeoFS Subnet"
+safemethods: ["version"]
+permissions:
+  - methods: ["update"]
--- a/subnet/subnet_contract.go
+++ b/subnet/subnet_contract.go
@@ -100,6 +100,26 @@ func Get(id []byte) []byte {
 	return raw.([]byte)
 }
 
+// Delete deletes subnet with the specified id.
+func Delete(id []byte) {
+	ctx := storage.GetContext()
+	key := append([]byte{ownerPrefix}, id...)
+	raw := storage.Get(ctx, key)
+	if raw == nil {
+		panic("delete:" + ErrNotExist)
+	}
+
+	owner := raw.([]byte)
+	if !runtime.CheckWitness(owner) {
+		panic("delete: owner witness check failed")
+	}
+
+	storage.Delete(ctx, key)
+
+	key[0] = infoPrefix
+	storage.Delete(ctx, key)
+}
+
 // Update method updates contract source code and manifest. Can be invoked
 // only by committee.
 func Update(script []byte, manifest []byte, data interface{}) {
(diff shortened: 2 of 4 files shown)


subnet/config.yml
```diff
@@ -0,0 +1,11 @@
...
+safemethods: ["version"]
+permissions:
+  - methods: ["update"]
+events:
+  - name: SubnetPut
+    parameters:
+      - name: ownerKey
+        type: PublicKey
+      - name: info
+        type: ByteArray
```
alexvanin, line 11: Go for new line

subnet/subnet_contract.go
```diff
@@ -0,0 +1,137 @@
...
+		id := common.InvokeID([]interface{}{ownerKey, info}, []byte("put"))
+		n := common.Vote(ctx, id, nodeKey)
+		if n < threshold {
+			return
+		}
+
+		common.RemoveVotes(ctx, id)
+	} else {
+		if !runtime.CheckWitness(ownerKey) {
+			panic("put: witness check failed")
```
alexvanin, line 73:
```suggestion
			panic("put: owner witness check failed")
```

subnet/subnet_contract.go
```diff
@@ -0,0 +1,137 @@
...
+		multiaddr := common.AlphabetAddress()
+		if !runtime.CheckWitness(multiaddr) {
+			panic("put: alphabet witness check failed")
+		}
+	}
+
+	stKey := append([]byte{ownerPrefix}, id...)
+	if storage.Get(ctx, stKey) != nil {
+		panic("put: " + ErrAlreadyExists)
+	}
```
alexvanin, line 85: Put this check before `if notaryDisable` condition. Early checks before `runtime.Notify` are appreciated by the clients and it saves a bit of wasted GAS in notary disabled environment.

subnet/subnet_contract.go
```diff
@@ -0,0 +1,137 @@
...
+
+	ctx := storage.GetContext()
+	notaryDisabled := storage.Get(ctx, notaryDisabledKey).(bool)
+	if notaryDisabled {
+		alphabet := common.AlphabetNodes()
+		nodeKey := common.InnerRingInvoker(alphabet)
+		if len(nodeKey) == 0 {
+			if !runtime.CheckWitness(ownerKey) {
+				panic("put: witness check failed")
```
carpawell, line 57:
```suggestion
				panic("put: owner witness check failed")
```

subnet/subnet_contract.go
```diff
@@ -0,0 +1,137 @@
...
+	storage.Put(ctx, stKey, info)
+}
+
+// Get returns info about subnet with the specified id.
+func Get(id []byte) []byte {
+	ctx := storage.GetContext()
+	key := append([]byte{infoPrefix}, id...)
+	raw := storage.Get(ctx, key)
+	if raw == nil {
+		panic(ErrNotExist)
```
carpawell, line 98: should it be panic when clients check if the network hasn't been deleted?

fyrchik: We can try to return `Null` but the we should also require non-empty info in `Put`, I think.

alexvanin (approved):

carpawell (approved):

subnet/subnet_contract.go
carpawell: just thought that `Null` or smth like that can be more user-friendly than checking the message of the contract panic (I just think that getting subnetwork can be the usual call, like an healthcheck of the network, to prove that the owner has not deleted the network)

but if panic is a common practice of returning errors from contracts, it's ok

PR #175, opened 2021-11-22
nns: allow to register arbitrary-level domains

### LINKED ISSUE #172: Allow register 3-level+ domains
One should be able to register any second level domain to existing domain.

fyrchik:
Close #172 .

So there are two problems which needs to be discussed:
1. I register a domain `a.com` and register an A record `y.z.a.com` for it. What should happen if I register `z.a.com` domain? Current behaviour: fail registration with the error message showing conflicting record.
2. I register a domain `y.z.com` and a record `x.y.z.com`. What happens with record `x.y.z.com`?
Current behaviour: it will fail to resolve.

### DIFF
--- a/nns/nns_contract.go
+++ b/nns/nns_contract.go
@@ -298,6 +298,17 @@ func Register(name string, owner interop.Hash160, email string, refresh, retry, 
 		nsBytes := storage.Get(ctx, append([]byte{prefixName}, parentKey...))
 		ns := std.Deserialize(nsBytes.([]byte)).(NameState)
 		ns.checkAdmin()
+
+		parentRecKey := append([]byte{prefixRecord}, parentKey...)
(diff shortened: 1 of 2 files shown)


realloc:
>I register a domain a.com and register an A record y.z.a.com for it. What should happen if I register z.a.com domain? Current behaviour: fail registration with the error message showing conflicting record.

It's ok to register z.a.com if there is NS record set for it in a.com zone. In NNS case there must be some permission from a.com zone owner.

> I register a domain y.z.com and a record x.y.z.com. What happens with record x.y.z.com?

When you create x.y.z.com zone, the x.y.z.com record becomes same as `@` record for newly created zone.

Maybe we should discuss usecases.

alexvanin (approved):

alexvanin:
Rebase, please.

PR #176, opened 2021-11-22
nns: add test for TLD record registration

### LINKED ISSUE #173: Top level domains can't store records
The one can't add any type of records into top level domain. This can be quite useful to store `SRV`, `TXT` or even `A` records in TLD. For example CDN domain can be mapped to `.cdn` TLD in NNS. Information about edge-managers can be encoded in `.cdn` records.

Based on @anatoly-bogatyrev feedback.

fyrchik:
Close #173.


alexvanin:
So it is possible to add records for TLD, right?

alexvanin (approved):

fyrchik:
Yes, the test is there.

PR #189, opened 2021-12-02
[#154] netmap: implement `Register` method

### LINKED ISSUE #154: Check both node and alphabet signature in netmap contract
Wait until nspcc-dev/neo-go#2150 is resolved.

fyrchik:
Close #154.

For notary-disabled environment it makes sense to split node
registration and actual candidate update into separate methods.
This way we have less complicate logic in `AddPeer` and overall
registration flow is more understandable.

### DIFF
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -188,6 +188,21 @@ func UpdateInnerRing(keys []interop.PublicKey) {
 	common.SetSerialized(ctx, innerRingKey, irList)
 }
 
+// Register method tries to add new candidate to the network map by
+// emitting AddPeer notification. Should be invoked by the registree.
+func Register(nodeInfo []byte) {
+	ctx := storage.GetContext()
+	notaryDisabled := storage.Get(ctx, notaryDisabledKey).(bool)
+	if notaryDisabled {
+		panic("Register should only be called in notary-enabled environment")
+	}
+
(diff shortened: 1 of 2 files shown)


alexvanin (commented):
With this change, storage nodes in notary enabled environment have to invoke `Register` method instead `AddPeer`. 

I thought it would be better to keep contract interface for storage nodes the same, and change only processing in Alphabet nodes. 

For example, `AddPeer` still produces notification and contains `common.CheckWitness(publicKey)`. While `Register` checks alphabet multi signature and adds `nodeInfo` to the storage. 

What do you think about that?

alexvanin:
> For notary-disabled environment it makes sense to split node

For notary-enabled, actually.

netmap/netmap_contract.go
```diff
@@ -188,6 +188,21 @@ func UpdateInnerRing(keys []interop.PublicKey) {
...
+	ctx := storage.GetContext()
+	notaryDisabled := storage.Get(ctx, notaryDisabledKey).(bool)
+	if notaryDisabled {
+		panic("Register should only be called in notary-enabled environment")
+	}
+
+	common.CheckAlphabetWitness(common.AlphabetAddress())
+
+	addToNetmap(ctx, storageNode{info: nodeInfo})
+	return
```
alexvanin, line 203: redundant return?

alexvanin (approved):

PR #195, opened 2021-12-08
Keep container contract as owner of `.container` zone

### LINKED ISSUE #194: Named container creation issue
In neofs-contract v0.13.0 I can't create named contract in notary disabled environment. I got 
```
could not approve put container {"error": "could not invoke method (putNamed): chain/client: contract execution finished with state FAULT; exception: at instruction 599 (THROW): unhandled exception: \"not witnessed by admin\""}
```

alexvanin:
Closes #194

### DIFF
--- a/nns/nns_contract.go
+++ b/nns/nns_contract.go
@@ -94,8 +94,34 @@ func _deploy(data interface{}, isUpdate bool) {
 		it := storage.Find(ctx, []byte{prefixRoot}, storage.KeysOnly|storage.RemovePrefix)
 		for iterator.Next(it) {
 			name := iterator.Value(it).(string)
+			if name != "neofs" {
+				continue
+			}
+
 			ns := NameState{
 				Owner:      committee,
+				Name:       name,
+				Expiration: runtime.GetTime() + millisecondsInYear,
+			}
+			tokenKey := getTokenKey([]byte(name))
+			putNameStateWithKey(ctx, tokenKey, ns)
+			putSoaRecord(ctx, name, "ops@nspcc.ru",
+				3600 /* 1 hour */, 600, /* 10 min */
+				604800 /* 1 week */, 3600 /* 1 hour */)
+		}
+
+		r := GetRecords("container.neofs", TXT)
+		owner := shBEFromLEHex(r[0])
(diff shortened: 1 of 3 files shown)


nns/nns_contract.go
```diff
@@ -113,6 +139,39 @@ func _deploy(data interface{}, isUpdate bool) {
...
+		res[ln-1-i] = n.([]byte)[0]
+	}
+
+	return res
+}
+
+// remove after v0.13.1 upgrade
+func hexToNum(s uint8) int {
+	if s >= '0' && s <= '9' {
+		return int(s) - 48
```
cthulhu-rider, line 161: For readability I'd write like this
```
return int(s) - '0'
...
return int(s) - 'A' + 10
...
return int(s) - 'a' + 10
```
but I'm not sure about the economy in terms of the contract

alexvanin: This is one time thing, it will be removed asap.

nns/nns_contract.go
```diff
@@ -113,6 +139,39 @@ func _deploy(data interface{}, isUpdate bool) {
 	storage.Put(ctx, []byte{prefixRegisterPrice}, defaultRegisterPrice)
 }
 
+// remove after v0.13.1 upgrade
+func shBEFromLEHex(s string) interop.Hash160 {
+	res := make([]byte, interop.Hash160Len)
+
+	ln := len(s) / 2
+	for i := 0; i < ln; i++ {
+		a := hexToNum(s[i*2])
```
cthulhu-rider, line 148:
```
// don't do ln := len(s) / 2
for i := 0; i < ln; i += 2 {
  a := hexToNum(s[i])
  b := hexToNum(s[i+1])
  ...
}
```
is cheaper isn't it?

cthulhu-rider (approved):

nns/nns_contract.go
alexvanin: I use `i` to address the byte in the resulting slice

```go
res[ln-1-i] = n.([]byte)[0]
```

In your example I will need to divide on each iteration to specify correct index of `res`.

PR #208, opened 2021-12-27
[#204] *: check contract version during update

### LINKED ISSUE #204: Process previous contract versions in update methods
After https://github.com/nspcc-dev/neofs-contract/pull/164 `Update` methods provide version of the contract to check it during update. Versions are provided but not processed. To process version, compare it with `common.PrevVersion` in `if isUpdate {...}` branch of `_deploy` function. Provided version should not be less than `common.PrevVersion`.

fyrchik:
Close #204.


alphabet/alphabet_contract.go
```diff
@@ -33,19 +33,21 @@ func OnNEP17Payment(from interop.Hash160, amount int, data interface{}) {
...
-	}
-
 	args := data.(struct {
 		notaryDisabled bool
 		addrNetmap     interop.Hash160
 		addrProxy      interop.Hash160
 		name           string
 		index          int
 		total          int
+		version        int
```
alexvanin, line 43: In initial deploy, `args` most likely won't have version field. Is it safe for VM to cast 6-element array into 7-element struct?

alexvanin: Also I can imagine the case when `data interface{}` during contract update would provide additional information which is not related to contract deploy arguments. 

I thought that the robust way to fetch version is to manually fetch the last element of `[]interface` in `isUpdate` branch. That is the only reliable property of `data interface{}` during contract update.

```go
if isUpdate {
  arr := args.([]interface{})
  ln := len(arr)
  var version int = arr[ln-1]
  common.CheckVersion(version)
}

Comment (fyrchik) [reply alphabet/alphabet_contract.go:43]: >Is it safe for VM to cast 6-element array into 7-element struct?

Yes, both array and struct types are unsized.

File: reputation/reputation_contract.go [hunk 2]
```diff
@@ -22,6 +22,9 @@ func _deploy(data interface{}, isUpdate bool) {
 	ctx := storage.GetContext()
 
 	if isUpdate {
+		args := data.([]interface{})
```
carpawell, line 25: why not cast to `struct`?

fyrchik: Se the comment above.

alexvanin (approved):

PR #213, opened 2022-01-13
Release/v0.14.0

alexvanin:
Neotest changes from #198 require neo-go version bump and this bump is incompatible with latest stable neo-go version. Other commits didn't require version bump so I reverted it until we will get new neo-go release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,22 @@
 # Changelog
 Changelog for NeoFS Contract
 
+## [0.14.0] - 2022-01-14 - Geojedo (거제도, 巨濟島)
+
+### Fixed
+- Sync `Update` method signature in NNS contract (#197)
+- Use current block index in all `GetDisgnatedByRole` invocations (#209)
+
+### Added
+- Version check during contract update (#204)
+
+### Changed
+- Use `storage.RemovePrefix` in subnet contract (#199)
+
+### Removed
+- Netmap contract hash usage in proxy contract (#205)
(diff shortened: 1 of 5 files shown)


alexvanin:
https://github.com/nspcc-dev/neofs-contract/issues/214

common/version.go
```diff
@@ -4,15 +4,15 @@ import "github.com/nspcc-dev/neo-go/pkg/interop/native/std"
...
+	patch = 0
 
 	// Versions from which an update should be performed.
 	// These should be used in a group (so prevMinor can be equal to minor if there are
 	// any migration routines.
 	prevMajor = 0
-	prevMinor = 12
-	prevPatch = 2
+	prevMinor = 13
+	prevPatch = 0
```
carpawell, line 15: should it be `0.13.0`?

carpawell (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,22 @@
 # Changelog
 Changelog for NeoFS Contract
 
+## [0.14.0] - 2022-01-13 - Geojedo (거제도, 巨濟島)
```
carpawell, line 4: date is no longer valid

common/version.go
alexvanin: Yes, we are upgrading from `v0.13.1` in N3 Testnet and `v0.13.2` in N3 Mainnet. Those versions are bigger than `v0.13.0`, so it should be okay.

PR #216, opened 2022-01-17
Make container size estimation removal more reliable

### LINKED ISSUE #215: Stuck of epoch updating process
During epoch update, `netmap` contract invokes clean up methods in `container` and `balance` contracts.

https://github.com/nspcc-dev/neofs-contract/blob/f5531561b86609ddfcad4f72313f5b11d2d0ded0/netmap/netmap_contract.go#L588-L594

- `balance` contract unlocks all expired accounts (if withdraw operation wasn't successful) 
- `container` contract removes old container size estimations

Many containers produce many container estimation records that should be removed on each epoch. Deleting those estimations in the same transaction with new epoch update will cause transaction to reach default RPC node invocation limit (15 GAS). Transaction won't be executed and new epoch will never happen. 

This also applies for many unlock transfers in `balance` contract (much less likely, though) and applies for general case when network map is too big to handle.

---

As the solution for big container amount, @fyrchik suggests to remove old estimation during `PutContainerSize` invocation.

Maybe RPC nodes of Inner Ring should have increased RPC invocation limit /cc @anatoly-bogatyrev @realloc 

One day epochs should be independent from network map changes and other corresponding operations like contract clean ups.

Thanks to @KirillovDenis for noting this issue.

fyrchik:
Close #215 .


tests/netmap_test.go
```diff
@@ -56,13 +56,28 @@ func TestDeploySetConfig(t *testing.T) {
...
+	pub    []byte
+	raw    []byte
+}
+
+func dummyNodeInfo(acc neotest.Signer) testNodeInfo {
 	ni := make([]byte, 66)
 	rand.Read(ni)
 
-	pub, _ := vm.ParseSignatureContract(acc.Script())
+	s := acc.(neotest.SingleSigner)
```
alexvanin, line 69: Requires neo-go version bump?

fyrchik: Actually no, we just forgot to fix this after update.

fyrchik: For some reason tests still fail, though work on my local machine. Updated neo-go version.

container/container_contract.go
```diff
@@ -728,6 +747,13 @@ func ownerFromBinaryContainer(container []byte) []byte {
 	return container[offset : offset+25] // offset + size of owner
 }
 
+func singleEstimationKey(epoch int, cid []byte, pub interop.PublicKey) []byte {
```
alexvanin, line 750: unused?

container/container_contract.go
```diff
@@ -774,9 +800,33 @@ func isStorageNode(ctx storage.Context, key interop.PublicKey) bool {
...
+func updateEstimations(ctx storage.Context, epoch int, cid []byte, pub interop.PublicKey, isUpdate bool) {
+	h := crypto.Ripemd160(pub)
+	estKey := append([]byte(singleEstimatePrefix), cid...)
+	estKey = append(estKey, h...)
+
+	var newEpochs []int
+	rawList := storage.Get(ctx, estKey).([]byte)
+
+	if rawList != nil {
+		epochs := std.Deserialize(rawList).([]int)
```
alexvanin, line 812: Storing slice is efficient, but won't be there any issues if multiple estimations with new epoch value are going to be in the same block? 

It shouldn't though, all tx will have enough GAS to invoke `storage.Delete` but only first tx in the block will do it, and others txs will be cheaper. But maybe I am missing something.

fyrchik: My initial thought was that because an estimate is produced by a certain node for a specific container, having multiple tx in one block is highly unlikely, given that epoch is much bigger than block time.
I wanted to get rid of `storage.Find` to reduce GAS cost (this will be amplified by iteration in `newEpoch`), but may be it doesn't worth the effort.

alexvanin (approved):

PR #217, opened 2022-01-18
Reverted #198

carpawell:
Same as previously reverted #198.


carpawell:
Think #216 should be merged first.

alexvanin (approved):
Rebase and merge

Issue #225, opened 2022-03-09
IR is not able to remove SN from netmap


carpawell:
After #162 it is impossible to make netmap cleanup by IR in notary-enabled mode: such `UpdateState` calls must be signed [both](https://github.com/nspcc-dev/neofs-contract/blob/v0.14.2/netmap/netmap_contract.go#L302) by a node and by the Alphabet.

IR still could process SN's requests to be removed but could not delete nodes that have not sent such requests (e.g. suddenly stopped): `netmap/process_cleanup.go:36 can't invoke netmap.UpdateState {"error": "could not invoke smart contract: chain/client: contract execution finished with state FAULT; exception: at instruction 1501 (THROW): unhandled exception: \"witness check failed\""} `

#### Possible solution:
1. Make two separate methods: for SN and for IR,
2. ...

alexvanin:
Seems similar to in https://github.com/nspcc-dev/neofs-contract/issues/154#issuecomment-970341435. We chose an option to split adding peer into two methods: one (`AddPeer`) with storage node witness check, another (`Register`) with alphabet multisig witness check.

I think we have to do the same thing there, split `UpdateState` into `UpdateState` with storage node witness check and `???` with alphabet multisig witness check.

PR #226, opened 2022-03-10
Increase snapshot history size

### LINKED ISSUE #224: Store more network map snapshots in netmap contract
According to [spec](https://github.com/nspcc-dev/neofs-api/blob/f6dad19bef26a049e05eaedfebf90852697c03a7/session/types.proto#L130-L138), Storage node should support `__NEOFS__NETMAP_EPOCH` and `__NEOFS__NETMAP_LOOKUP_DEPTH` x-headers for object service requests. 

It is not clear how we should implement it, because Storage nodes do not cache network maps and netmap contract contains only one previous epoch.

fyrchik:
Close #224 , depends on #225 .

### DIFF
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -37,7 +37,7 @@ const (
 
 	// SnapshotCount contains the number of previous snapshots stored by this contract.
 	// Must be less than 255.
-	SnapshotCount        = 2
+	SnapshotCount        = 10
 	snapshotKeyPrefix    = "snapshot_"
 	snapshotCurrentIDKey = "snapshotCurrent"
 	snapshotEpoch        = "snapshotEpoch"
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -35,10 +35,13 @@ const (
 	notaryDisabledKey = "notary"
 	innerRingKey      = "innerring"
 
-	snapshot0Key     = "snapshotCurrent"
-	snapshot1Key     = "snapshotPrevious"
-	snapshotEpoch    = "snapshotEpoch"
-	snapshotBlockKey = "snapshotBlock"
+	// SnapshotCount contains the number of previous snapshots stored by this contract.
+	// Must be less than 255.
+	SnapshotCount        = 2
+	snapshotKeyPrefix    = "snapshot_"
+	snapshotCurrentIDKey = "snapshotCurrent"
+	snapshotEpoch        = "snapshotEpoch"
+	snapshotBlockKey     = "snapshotBlock"
 
 	containerContractKey = "containerScriptHash"
 	balanceContractKey   = "balanceScriptHash"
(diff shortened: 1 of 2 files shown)


netmap/netmap_contract.go
```diff
@@ -35,10 +35,13 @@ const (
 	notaryDisabledKey = "notary"
 	innerRingKey      = "innerring"
 
-	snapshot0Key     = "snapshotCurrent"
-	snapshot1Key     = "snapshotPrevious"
-	snapshotEpoch    = "snapshotEpoch"
-	snapshotBlockKey = "snapshotBlock"
+	// SnapshotCount contains the number of previous snapshots stored by this contract.
+	// Must be less than 255.
+	SnapshotCount        = 10
```
carpawell, line 40: - may it be useful to be able to change that value?
- may it be useful to be able to get that value from the contract?

fyrchik: Both yes.

netmap/netmap_contract.go
```diff
@@ -396,7 +430,7 @@ func LastEpochBlock() int {
 // of current epoch.
 func Netmap() []storageNode {
 	ctx := storage.GetReadOnlyContext()
-	return getSnapshot(ctx, snapshot0Key)
+	return getSnapshot(ctx, snapshotKeyPrefix+"0")
```
alexvanin, line 433: Shouldn't it be `snapshotKeyPreifx+storage.Get(snapshotCurrentIDKey)`?

netmap/netmap_contract.go
```diff
@@ -414,18 +448,14 @@ func NetmapCandidates() []netmapNode {
...
 func Snapshot(diff int) []storageNode {
-	var key string
-
-	switch diff {
-	case 0:
-		key = snapshot0Key
-	case 1:
-		key = snapshot1Key
-	default:
+	if SnapshotCount <= diff {
```
alexvanin, line 451: Negative `diff` should be invalid too.

netmap/netmap_contract.go
```diff
@@ -85,6 +88,15 @@ func _deploy(data interface{}, isUpdate bool) {
 
 	if isUpdate {
 		common.CheckVersion(args.version)
+
+		data := storage.Get(ctx, "snapshotCurrent")
+		storage.Put(ctx, snapshotKeyPrefix+"0", data)
+
+		data = storage.Get(ctx, "snapshotPrevious")
+		storage.Put(ctx, snapshotKeyPrefix+"1", data)
```
alexvanin, line 96: If I understand correctly, we put new epochs with incremented `id`. So previous snapshot should be with `id:0`, current snapshot with `id:1` and `snapshotCurrentIDKey:1`.

netmap/netmap_contract.go
fyrchik: Fixed.

netmap/netmap_contract.go
fyrchik: Fixed.

netmap/netmap_contract.go
fyrchik: Yes, fixed.

alexvanin (approved):

PR #227, opened 2022-03-10
netmap: split `UpdateState` method for notary-enabled environment

fyrchik:
Split methods similar to `AddPeer`/`Register` pair from b104a2ccbcfdb .


alexvanin:
> The one with Peer in name should be called by a storage node similar to AddPeer/Register.

While it is better naming scheme for understanding, I am afraid it will break compatibility with older storage nodes as they continue to invoke `UpdateState` instead of `UpdatePeerState`.

netmap/netmap_contract.go
```diff
@@ -281,13 +281,17 @@ func UpdateState(state int, publicKey interop.PublicKey) {
...
+	}
 
-			runtime.Notify("UpdateState", state, publicKey)
-			return
-		}
+	// If notary is enabled or caller is not an alphabet node,
+	// just emit the notification for alphabet.
+	if !notaryDisabled || len(nodeKey) == 0 {
+		common.CheckWitness(publicKey)
+		runtime.Notify("UpdateState", state, publicKey)
```
carpawell, line 290: why have you decided to produce a notification in notary-enabled mode?

netmap/netmap_contract.go
```diff
@@ -190,9 +190,9 @@ func UpdateInnerRing(keys []interop.PublicKey) {
 	common.SetSerialized(ctx, innerRingKey, irList)
 }
 
-// Register method tries to add new candidate to the network map by
-// emitting AddPeer notification. Should be invoked by the registree.
-func Register(nodeInfo []byte) {
+// AddPeerIR method tries to add new candidate to the network map.
+// Should be invoked in notary-enabled environment by the alphabet.
```
carpawell, line 194: add somehow info that this should be invoked "*only* in notary-enabled...", please

netmap/netmap_contract.go
```diff
@@ -312,6 +312,25 @@ func UpdateState(state int, publicKey interop.PublicKey) {
 	}
 }
 
+// UpdateStateIR method tries to change node state in the network map.
+// Should be invoked in notary-enabled environment by the alphabet.
```
carpawell, line 316: add somehow info that this should be invoked "*only* in notary-enabled...", please

netmap/netmap_contract.go
fyrchik: To mirror `AddPeer` behaviour.

netmap/netmap_contract.go
carpawell: i meant that it could be read as "Should be invoked ... by the alphabet", not clear that it is prohibited to use in notary disabled environment at all

netmap/netmap_contract.go
fyrchik: Fixed.

netmap/netmap_contract.go
fyrchik: Fixed.

netmap/netmap_contract.go
alexvanin: Agree that those notifications in notary enabled environment are kinda useless. Let's remove them and put useful notifications directly to `removeFromNetmap` and `addToNetmap` functions.

fyrchik: Fixed, removed.

alexvanin (approved):

PR #228, opened 2022-03-21
Remove IRNode from sidechain contracts

### LINKED ISSUE #222: Replace `common.IRNode` with `interop.PublicKey`
This is a structure with a single field, we can save on contract size and runtime GAS cost if we use `interop.PublicKey` directly.

fyrchik:
Close #222.

This leads to simpler code and lower GAS cost for transactions.
Also do some other optimizations.

### DIFF
--- a/common/ir.go
+++ b/common/ir.go
@@ -17,11 +17,11 @@ const irListMethod = "innerRingList"
 
 // InnerRingInvoker returns public key of inner ring node that invoked contract.
 // Work around for environments without notary support.
-func InnerRingInvoker(ir []IRNode) interop.PublicKey {
+func InnerRingInvoker(ir []interop.PublicKey) interop.PublicKey {
 	for i := 0; i < len(ir); i++ {
 		node := ir[i]
-		if runtime.CheckWitness(node.PublicKey) {
-			return node.PublicKey
+		if runtime.CheckWitness(node) {
+			return node
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


netmap/netmap_contract.go
```diff
@@ -137,7 +146,7 @@ func Update(script []byte, manifest []byte, data interface{}) {
 //
 // If notary enabled, then look to NeoFSAlphabet role in native RoleManagement
 // contract of the side chain.
-func InnerRingList() []common.IRNode {
+func InnerRingList() []interop.PublicKey {
```
alexvanin, line 149: While this public method is still a thing (until notary will be enabled in the chains), I propose to keep return value the same.

common/ir.go
```diff
@@ -30,23 +30,21 @@ func InnerRingInvoker(ir []IRNode) interop.PublicKey {
...
+	return roles.GetDesignatedByRole(roles.NeoFSAlphabet, uint32(blockHeight+1))
 }
 
 // InnerRingNodesFromNetmap gets list of inner ring through
 // calling "innerRingList" method of smart contract.
 // Work around for environments without notary support.
-func InnerRingNodesFromNetmap(sc interop.Hash160) []IRNode {
-	return contract.Call(sc, irListMethod, contract.ReadOnly).([]IRNode)
+func InnerRingNodesFromNetmap(sc interop.Hash160) []interop.PublicKey {
+	return contract.Call(sc, irListMethod, contract.ReadOnly).([]interop.PublicKey)
```
alexvanin, line 42: After latest change, `irListMethod` returns `[]IRNode`. So we had to convert it into `[]interop.Public.Keys` inside of this method.

Fortunately, it has only two references in whole repository, so it won't be a big overhead. Also `InnerRingNodesFromNetmap` will be removed after enabling notary subsystem.

fyrchik: Fixed.

alexvanin (approved):

PR #233, opened 2022-03-28
Configure netmap snapshot history size

### LINKED ISSUE #232: Configure amount of stored network maps
Provide method to change amount of stored network maps in the contract.

fyrchik:
Close #232 .

### DIFF
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -35,9 +35,10 @@ const (
 	notaryDisabledKey = "notary"
 	innerRingKey      = "innerring"
 
-	// SnapshotCount contains the number of previous snapshots stored by this contract.
+	// DefaultSnapshotCount contains the number of previous snapshots stored by this contract.
 	// Must be less than 255.
-	SnapshotCount        = 10
+	DefaultSnapshotCount = 10
+	snapshotCountKey     = "snapshotCount"
 	snapshotKeyPrefix    = "snapshot_"
 	snapshotCurrentIDKey = "snapshotCurrent"
 	snapshotEpoch        = "snapshotEpoch"
@@ -88,6 +89,7 @@ func _deploy(data interface{}, isUpdate bool) {
 
 	if isUpdate {
 		common.CheckVersion(args.version)
+		storage.Put(ctx, snapshotCountKey, DefaultSnapshotCount)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


netmap/netmap_contract.go
```diff
@@ -457,17 +441,101 @@ func NetmapCandidates() []netmapNode {
...
+		// EPOCH    : C-E | C-E+1 | ... | C | nil | ... | .   | ... | C-E-1
+		//
+		// So we need to move tail snapshots N-K keys forward,
+		// i.e. from E+1 .. K to N-K+E+1 .. N
+		diff := count - curr
+		lower := diff + id + 1
+		for k := count - 1; k >= lower; k-- {
+			moveSnapshot(ctx, k-diff, k)
+		}
+		delStart, delFinish = id+1, id+1+diff
```
alexvanin, line 497: Maybe `delFinish` can be `curr`?

Consider expanding 3 to 6 with `id` as last element:
```
before: [a, b, c]
               ^-id
after:  [a, b, c, 0, 0, 0]  
               ^-id
```

In current code `delStart = 3` and `delFinish=6` but actually you don't need to remove any values there.

netmap/netmap_contract.go
```diff
@@ -457,17 +441,101 @@ func NetmapCandidates() []netmapNode {
...
+		// KEY INDEX: 0   | 1     | ... K1 ... | E | E+1   | ... K2-1 ... | N-1
+		// EPOCH    : C-E | C-E+1 | ... .. ... | C | C-N+1 | ... ...  ... | C-E-1
+		var step, start int
+		if id < count {
+			// K2 case, move snapshots from E+1+N-K .. N-1 range to E+1 .. K-1
+			// New state:
+			// KEY INDEX: 0   | 1     | ... | E | E+1   | ... | K-1
+			// EPOCH    : C-E | C-E+1 | ... | C | C-K+1 | ... | C-E-1
+			step = curr - count
+			start = step + id + 1
```
alexvanin, line 511: I think it should be `start = id + 1` there. You get `+step` in the `for` loop down below. 

In current implementation with this example I get `step = 2`, `start = 5`, therefore `for` loop has 0 iteration.
```
before: [g, h, i, d, e, f]
               ^ - id
after:  [g, h, i, f]  
               ^ - id
```

netmap/netmap_contract.go
```diff
@@ -409,7 +393,7 @@ func NewEpoch(epochNum int) {
 	storage.Put(ctx, snapshotBlockKey, ledger.CurrentIndex())
 
 	id := storage.Get(ctx, snapshotCurrentIDKey).(int)
-	id = (id + 1) % SnapshotCount
+	id = (id + 1) % DefaultSnapshotCount
```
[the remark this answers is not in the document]
fyrchik: Fixed

netmap/netmap_contract.go
fyrchik: True, fixed.

fyrchik: Actually, no `id+1` isn't valid if count difference is less than `current count - current id`. 
I think it should be something like `min(id+1+diff, curr)`.

alexvanin (approved):

PR #237, opened 2022-04-01
[#236] *: Enable completion notifications for notary-disabled environ…

### LINKED ISSUE #236: Add notifications of successfull operations
Notary-disabled environment produces notifications to initiate voting mechanism. However there is (almost) no notification to notify about operation success. It can be very useful for awaiting operations (such as awaiting for new container to show up) and for monitoring the whole network.

I suggest to add such notifications. Some useful events:
- container has been created
- container has been deleted
- container has changed eACL
- storage node has been registered as network map candidate
- storage node has changed its state

Notifications must be described in `doc.go` files.

This should be relevant after enabling notary. We can try to implement it already, but there might be some collision in notification names with notary-disabled notifications.

fyrchik:
…ment

Close #236 


netmap/netmap_contract.go
```diff
@@ -256,6 +256,9 @@ func AddPeer(nodeInfo []byte) {
 	}
 
 	addToNetmap(ctx, candidate)
+	if notaryDisabled {
```
alexvanin, line 259: Success notifications are useful for both notary enabled and disabled environments, no?

netmap/netmap_contract.go
```diff
@@ -316,6 +317,8 @@ func UpdateState(state int, publicKey interop.PublicKey) {
 	default:
 		panic("unsupported state")
 	}
+
+	runtime.Notify("UpdateStateSuccess", state, publicKey)
```
alexvanin, line 321: `UpdateStateIR` method should produce the same notification, I think.

netmap/netmap_contract.go
```diff
@@ -256,6 +256,7 @@ func AddPeer(nodeInfo []byte) {
 	}
 
 	addToNetmap(ctx, candidate)
+	runtime.Notify("AddPeerSuccess", nodeInfo)
```
alexvanin, line 259: `AddPeerIR` method too?

container/container_contract.go
```diff
@@ -357,6 +358,7 @@ func Delete(containerID []byte, signature interop.Signature, token []byte) {
 	}
 	removeContainer(ctx, containerID, ownerID)
 	runtime.Log("remove container")
+	runtime.Notify("DeleteSuccess", interop.Hash256(containerID), signature, token)
```
alexvanin, line 361: Why hash container ID?

alexvanin (commented):
Some changes in `doc.go` will be appreciated. They are go into specification pdf.

container/container_contract.go
fyrchik: It is a type case, to have proper argument type in the manifest.

netmap/netmap_contract.go
fyrchik: Done

netmap/netmap_contract.go
```diff
@@ -202,6 +202,9 @@ func AddPeerIR(nodeInfo []byte) {
 	common.CheckAlphabetWitness(common.AlphabetAddress())
 
 	addToNetmap(ctx, storageNode{info: nodeInfo})
+
+	publicKey := nodeInfo[2:35]
```
alexvanin, line 206: Comment about `// V2 format` to explain those magic nums?

alexvanin (approved):

PR #250, opened 2022-07-18
Makefile: allow to build from the source archive

fyrchik:


alexvanin:
Good first issue: fetch contract version from `common/versions.go` by using awk instated of VERSION file :smile:

PR #265, opened 2022-09-06
Adjust container-contract permissions

AnnaShaleva:
Remove "addRoot" method from the list of allowed methods, because
NNS doesn't have it since
https://github.com/nspcc-dev/neofs-contract/pull/139/commits/4b86891d57ef12c3f37a9ef27902c73d96812934.


carpawell (approved):
@AnnaShaleva, could you, please, change the scope in the commit message: "neofs" -> "container". We have "neofs" contract too.

fyrchik (approved):

PR #266, opened 2022-09-09
Update NNS contract

AnnaShaleva:
Refactor code and fix a couple of bugs. Thes PR may affect other services, so @fyrchik, @carpawell, @alexvanin please, review it carefully.

@roman-khimov, these changes are ported to https://github.com/nspcc-dev/neo-go/pull/2687 and to the set of C# PRs (https://github.com/neo-project/non-native-contracts/pull/27 and others).


nns/nns.yml
```diff
@@ -1,7 +1,7 @@
 name: "NameService"
 supportedstandards: ["NEP-11"]
 safemethods: ["balanceOf", "decimals", "symbol", "totalSupply", "tokensOf", "ownerOf",
-              "tokens", "properties", "roots", "getPrice", "isAvailable", "getRecord",
+              "tokens", "properties", "roots", "getPrice", "isAvailable", "getRecords",
```
roman-khimov, line 4: Hmm. I'd expect our compiler to catch missing `getRecord`.

nns/nns_contract.go
```diff
@@ -452,11 +452,8 @@ func Resolve(name string, typ RecordType) []string {
 
 // GetAllRecords returns an Iterator with RecordState items for the given name.
 func GetAllRecords(name string) iterator.Iterator {
-	tokenID := []byte(tokenIDFromName(name))
 	ctx := storage.GetReadOnlyContext()
-	_ = getNameState(ctx, tokenID) // ensure not expired
-	recordsKey := getRecordsKey(tokenID, name)
-	return storage.Find(ctx, recordsKey, storage.ValuesOnly|storage.DeserializeValues)
+	return getAllRecords(ctx, name)
```
roman-khimov, line 456: If we have a `maxRecordID` and a limited set of types, maybe we can avoid having `iterator.Iterator`? Although we still should have some reasonable number of items in the end. Just an idea.

roman-khimov (commented):
Seems to be reasonable.

nns/nns.yml
AnnaShaleva: And it will. But I found out that `nns.yml` is unused config (there's `config.yml` instead), so I'll remove it.

nns/nns_contract.go
AnnaShaleva: Maybe we firstly need to discuss `maxRecordID` value? I've added this constraint because of the record ID type: it was `byte`, but all int-like go types are `BiGInteger` in the VM, so technically, the previous implementation didn't restrict the maximum number of records of the same type. I suppose, it's not the initial intention, thus let's decide on `maxRecordID`.

And if we're to keep `MaxRecordID` set to be 255, then I vote for converting the result of `getAllecords` to `[]stirng`.

roman-khimov: DNS doesn't have this limit. C# NNS doesn't have it too. But I see what you're talking about now, `id byte` overflow can be a problem in some cases. It can be fixed at the same time to have unlimited number of records of any type. Practically I'd rather have some limit (16-64), at the moment C# contract has a limit of 1 IIRC and it's kinda OK.

fyrchik:
@AnnaShaleva we agreed to use capital letters in NeoFS repos after `nns: ` part, could you change this? You can use my script if you want https://gist.github.com/fyrchik/0c7075d03f4bfe3a7665e98f9100efdb (`beautify #issue master`)

nns/nns_contract.go
AnnaShaleva: @fyrchik, what do you think about it?

fyrchik: I think `[]string` is easier to work with. In our scenarios we have at most 2 records per type, so 255 is more than enough.

AnnaShaleva: Getting back to this discussion, the profit of `Iterator` returned value here is that it returns not only record name, but the complete record information which includes record's name,  type, ID and data. Currently it's the only method allowing to retrieve the whole `RecordState` from the contract, so I've changed my mind and suggest to keep `Iterator`. @roman-khimov, somehow I've missed it previously.

AnnaShaleva:
@fyrchik, @roman-khimov need to be reviewed one more time, new commits are added.

carpawell:
Hey, what's up here? The anniversary coming up soon.

roman-khimov:
> The anniversary coming up soon

Preparing for celebration, as usual.

carpawell:
So are the conflicts gonna be solved? Is the PR's idea actual? Haven't seen the code yet, should I review this?

roman-khimov:
This one was superseded by #419 and #420.

PR #267, opened 2022-09-15
NameService: port changes from C# NNS

AnnaShaleva:
Depends on #266, will be rebased after upstream merge.

This PR includes changes implemented into C# NNS version (https://github.com/neo-project/non-native-contracts/blob/master/src/NameService/NameService.cs):

1. Domain registration price is dependant on domain name length. By default, short (<5 chars) domain names are available for registration only for committee.
    - [ ] This change may affect other neofs services including neofs-adm, so we need to create corresponding issues in the affected repos. 
    - [ ] Also, the contract storage scheme is changed, so the migration routine is added and we need to create an issue to remove it after the contract update.

2. `Renew` overload is added that accepts expiration period as an argument. This change is backwords-compatible.
3. `SetAdmin` and `Renew` events are added.
    - [ ] These events may be useful for neofs services, we may create an issue to use them where possible (if needed).

I'm restricted to choose the reviewers via Github interface, thus @roman-khimov, @fyrchik, review, please.


roman-khimov:
Replaced by #442.

PR #271, opened 2022-09-20
Support maintenance mode

### LINKED ISSUE #269: netmap: Support `MAINTENANCE` node state
* based on https://github.com/nspcc-dev/neofs-api/issues/237
* blocked by feature release, most likely https://github.com/nspcc-dev/neofs-api/milestone/19

There is a need to support one more value of node state enum - `MAINTENANCE`. In addition to adding a new constant, you need to look at how specific values are processed in contract methods.

### LINKED ISSUE #270: netmap: Can we export the `nodeState` type?
It'd be useful to use this type for implementation of the Storage Node and Inner Ring applications. Enum elements are exported, but the type is not. Is there any reason for this, or is it free to export?

fyrchik:
Close #269.
Close #270.

### DIFF
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -56,6 +56,7 @@ const (
 	_ NodeState = iota
 	OnlineState
 	OfflineState
+	MaintenanceState
 )
 
 var (
@@ -317,6 +318,9 @@ func UpdateState(state int, publicKey interop.PublicKey) {
 	case OfflineState:
 		removeFromNetmap(ctx, publicKey)
 		runtime.Log("remove storage node from the network map")
+	case MaintenanceState:
+		updateNetmapState(ctx, publicKey, MaintenanceState)
+		runtime.Log("move storage node to a maintenance state")
 	default:
 		panic("unsupported state")
 	}
@@ -338,6 +342,8 @@ func UpdateStateIR(state NodeState, publicKey interop.PublicKey) {
 	switch state {
 	case OfflineState:
 		removeFromNetmap(ctx, publicKey)
+	case MaintenanceState:
+		updateNetmapState(ctx, publicKey, MaintenanceState)
 	default:
 		panic("unsupported state")
 	}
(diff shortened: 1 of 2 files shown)


netmap/netmap_contract.go
```diff
@@ -52,9 +55,10 @@ const (
 
 const (
 	// V2 format
```
carpawell, line 57: hm, that is not a `V2 format` then if "For historical reasons numerical values differs from the ones described in the NeoFS API."?

netmap/netmap_contract.go
```diff
@@ -20,10 +20,13 @@ type (
 
 	netmapNode struct {
 		node  storageNode
-		state nodeState
+		state NodeState
 	}
 
-	nodeState int
+	// NodeState is an enumeration for node states.
+	// NOTE: For historical reasons numerical values differs
```
carpawell, line 27:
```suggestion
	// NOTE: For historical reasons numerical values differ
```

netmap/netmap_contract.go
```diff
@@ -312,10 +316,13 @@ func UpdateState(state int, publicKey interop.PublicKey) {
 		common.CheckAlphabetWitness(common.AlphabetAddress())
 	}
 
-	switch nodeState(state) {
+	switch NodeState(state) {
 	case OfflineState:
 		removeFromNetmap(ctx, publicKey)
 		runtime.Log("remove storage node from the network map")
+	case MaintenanceState:
```
carpawell, line 323: previously we use that method only for removing from netmap since Online == `AddPeer`

now we have a new scenario: `Maintenance` => `Online`. won't it fail with `panic("unsupported state")`? think doesn't matter whether I'm right or not, we could also add that case in our tests since it gonna be a regular case

netmap/netmap_contract.go
fyrchik: Actually, it is my mistake, looks like they are the same, just that we had some logic in IR node.

netmap/netmap_contract.go
fyrchik: Fixed

cthulhu-rider (approved):

carpawell (approved):

alexvanin (approved):

PR #273, opened 2022-09-28
netmap: Return same type from reading methods

cthulhu-rider:
* relates to #269

I've changed format of the Netmap contract storage - should I do some compatibility actions? @fyrchik

### DIFF
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -13,24 +13,35 @@ import (
 	"github.com/nspcc-dev/neofs-contract/common"
 )
 
-type (
-	storageNode struct {
-		info []byte
-	}
+// NodeState is an enumeration for node states.
+type NodeState int
 
-	netmapNode struct {
-		node  storageNode
-		state NodeState
-	}
+// Various Node states
+const (
+	_ NodeState = iota
 
-	// NodeState is an enumeration for node states.
-	NodeState int
+	// NodeStateOnline stands for nodes that are in full network and
+	// operational availability.
+	NodeStateOnline
 
-	record struct {
-		key []byte
-		val []byte
-	}
+	// NodeStateOffline stands for nodes that are in network unavailability.
+	NodeStateOffline
+
+	// NodeStateMaintenance stands for nodes under maintenance with partial
+	// network availability.
+	NodeStateMaintenance
 )
+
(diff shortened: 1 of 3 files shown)


netmap/netmap_contract.go
```diff
@@ -13,25 +13,36 @@ import (
...
+	NodeStateOffline
+
+	// NodeStateMaintenance stands for nodes under maintenance with partial
+	// network availability.
+	NodeStateMaintenance
 )
 
+// Node groups data related to NeoFS storage nodes registered in the NeoFS
+// network. The information is stored in the current contract.
+type Node struct {
```
AnnaShaleva, line 37: If you're planning to update contract in the existing chain, then you need to write a migration routine. It should migrate the old netmap contract storage scheme to the new one (storageNode -> Node transformation), otherwise contract won't be able to work with the previously saved data.

netmap/netmap_contract.go
```diff
@@ -259,28 +287,63 @@ func AddPeer(nodeInfo []byte) {
...
+
+// updates state of the network map candidate by its public key in the contract
+// storage, and throws UpdateStateSuccess notification after this.
+//
+// State MUST be from the NodeState enum.
+func updateCandidateState(ctx storage.Context, publicKey interop.PublicKey, state NodeState) {
+	switch state {
+	case NodeStateOffline:
+		removeFromNetmap(ctx, publicKey)
+		runtime.Log("remove storage node from the network map")
```
AnnaShaleva, line 301: Why do you need these logs? There's a notification in the end of `updateCandidateState`, it should be enough.

fyrchik (commented):
LGTM, except for migration

netmap/netmap_contract.go
cthulhu-rider: Maybe to distingush notification losses from non-throws. I don't mind to have such log messages.

netmap/netmap_contract.go
cthulhu-rider: Yes, migration job should be performed, thanks.

I've started from documenting (for developers)  the storage structure #276.

cthulhu-rider: We've decided to postpone the migration #277.

fyrchik (approved):

PR #278, opened 2022-09-30
netmap: migrate from the old storage format

fyrchik:
Need this for a release.
Tested on neofs-adm from https://github.com/nspcc-dev/neofs-node/pull/1903

### DIFF
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -94,6 +94,25 @@ func _deploy(data interface{}, isUpdate bool) {
 
 	if isUpdate {
 		common.CheckVersion(args.version)
+
+		count := getSnapshotCount(ctx)
+		prefix := []byte(snapshotKeyPrefix)
+		for i := 0; i < count; i++ {
+			key := append(prefix, byte(i))
+			data := storage.Get(ctx, key)
+			if data != nil {
+				nodes := std.Deserialize(data.([]byte)).([]Node)
+				for i := range nodes {
+					// Old structure contains only the first field,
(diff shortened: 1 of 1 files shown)


carpawell (dismissed):

netmap/netmap_contract.go
```diff
@@ -94,6 +94,24 @@ func _deploy(data interface{}, isUpdate bool) {
 
 	if isUpdate {
 		common.CheckVersion(args.version)
+
+		prefix := []byte(snapshotKeyPrefix)
+		for i := 0; i < DefaultSnapshotCount; i++ {
```
cthulhu-rider, line 99: Shouldn't we read snapshot count from `snapshotCountKey`?

netmap/netmap_contract.go
```diff
@@ -94,6 +94,24 @@ func _deploy(data interface{}, isUpdate bool) {
...
+			key := append(prefix, byte(i))
+			data := storage.Get(ctx, key)
+			if data != nil {
+				nodes := std.Deserialize(data.([]byte)).([]Node)
```
cthulhu-rider, line 103: Can we decode to list of structures with old format in order to not loose the state?

fyrchik: I think it is ok now -- the only new state is maintenance and it is added in this release, so all ready stored nodes are `Online`.

netmap/netmap_contract.go
fyrchik: Fixed

cthulhu-rider (approved):

PR #288, opened 2022-10-17
[#285] Makefile: Install neo-go in the current directory

### LINKED ISSUE #285: `cli` artefact is left after almost all the `make` commands
Every build `make` command requires a `neo-go` binary. It is solved by [installing](https://github.com/nspcc-dev/neofs-contract/blob/6398b16a531d5f1bc128bfc0c98befea479230d3/Makefile#L38-L40) it as a `cli` global binary. IMO, it confuses and is not expected when building contracts, especially if none of the "clean" commands removes it.

Ideas:
1. Install/download `neo-go` locally in the repo dir.
2. Do not install `neo-go` if it is already installed.
3. Install `neo-go` anyway, but with proper naming and functionality (the current `cli` does not know what it is and what version it has).
4. ...

@cthulhu-rider, @fyrchik, @acid-ant

fyrchik:
Close #285.

`go install` doesn't have `-o` flag but it respects `GOBIN` environment variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 #!/usr/bin/make -f
 
 SHELL=bash
-GOBIN ?= $(shell go env GOPATH)/bin
+# GOBIN is used only to install neo-go and allows to override
+# the location of written binary.
+export GOBIN ?= $(shell pwd)/bin
 NEOGO ?= $(GOBIN)/cli
 VERSION ?= $(shell git describe --tags --dirty --match "v*" --always --abbrev=8 2>/dev/null || cat VERSION 2>/dev/null || echo "develop")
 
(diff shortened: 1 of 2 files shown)


cthulhu-rider (approved):

Makefile
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/make -f
 
 SHELL=bash
-GOBIN ?= $(shell go env GOPATH)/bin
+# GOBIN is used only to install neo-go and allows to override
+# the location of written binary.
+export GOBIN ?= $(shell pwd)/bin
```
carpawell, line 6: does `CURDIR` make the same?

carpawell (approved):

Makefile
fyrchik: It does, but it is a GNU extension, I am not sure where if this can bite us.

Issue #293, opened 2022-11-17
Implement method to iterate over estimated container sizes


vdomnich-yadro:
To process all container size estimates on epoch we are using combination of 2 methods of container smart contract:
* [ListContainerSizes](https://github.com/nspcc-dev/neofs-contract/blob/support/v0.16/container/container_contract.go#L555)
* [GetContainerSize](https://github.com/nspcc-dev/neofs-contract/blob/support/v0.16/container/container_contract.go#L541)

However, when there are many containers in the system, this method doesn't work efficiently. It takes about 3-4 seconds to process estimates of ~500 containers. To speed up processing on a system with multiple containers we want to implement an alternative method that would have signature:
```
func IterateContainerSizes(epoch int) iterator.Iterator { }
```

roman-khimov:
Fixed by #295.

PR #295, opened 2022-11-17
[#293] container: Add IterateContainerSizes method

vdomnich-yadro:
Add method that allows to iterate over estimation records.

Update tests to assert that list of estimations built with existing methods is identical to estimations from iterator.

### DIFF
--- a/container/container_contract.go
+++ b/container/container_contract.go
@@ -551,7 +551,7 @@ func GetContainerSize(id []byte) containerSizes {
 }
 
 // ListContainerSizes method returns the IDs of container size estimations
-// that has been registered for the specified epoch.
+// that have been registered for the specified epoch.
 func ListContainerSizes(epoch int) [][]byte {
 	ctx := storage.GetReadOnlyContext()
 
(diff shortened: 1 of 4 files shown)


container/container_contract.go
```diff
@@ -582,6 +582,19 @@ func ListContainerSizes(epoch int) [][]byte {
 	return result
 }
 
+// IterateContainerSizes method returns iterator over container size estimations
+// that have been registered for the specified epoch.
+func IterateContainerSizes(epoch int) iterator.Iterator {
```
fyrchik, line 587: Do we lose the information about the container in the result iterator? It seems just a key+size pair.

container/container_contract.go
```diff
@@ -551,7 +551,7 @@ func GetContainerSize(id []byte) containerSizes {
 }
 
 // ListContainerSizes method returns the IDs of container size estimations
-// that has been registered for the specified epoch.
+// that have been registered for the specified epoch.
```
fyrchik, line 554: Seems unrelated to this commit.

container/container_contract.go
fyrchik: Key is the key of a storage node, I doubt we can restore container ID from it.

fyrchik: I think I don't understand the use-case. Will it be sufficient to just return total size for the epoch?
If not, what is done with the values obtained via this method?

alexvanin: I think we should return both key and value, so container client will define iteration algorithm according to its needs.
As for total size, I agree that separate method will be sufficient.

fyrchik (approved):

alexvanin (approved):
@fyrchik Are we going to merge it in support branch or master? Right now target is support/v0.16

PR #300, opened 2022-12-22
container: allow to iterate over container list

fyrchik:

### DIFF
--- a/container/container_contract.go
+++ b/container/container_contract.go
@@ -62,6 +62,8 @@ const (
 
 	singleEstimatePrefix = "est"
 	estimateKeyPrefix    = "cnr"
+	containerKeyPrefix   = 'x'
+	ownerKeyPrefix       = 'o'
 	estimatePostfixSize  = 10
 	// CleanupDelta contains the number of the last epochs for which container estimations are present.
 	CleanupDelta = 3
@@ -93,6 +95,26 @@ func _deploy(data interface{}, isUpdate bool) {
 	if isUpdate {
 		args := data.([]interface{})
 		common.CheckVersion(args[len(args)-1].(int))
+
+		it := storage.Find(ctx, []byte{}, storage.None)
+		for iterator.Next(it) {
+			item := iterator.Value(it).(struct {
+				key   []byte
+				value []byte
+			})
+
+			// Migrate container.
+			if len(item.key) == containerIDSize {
+				storage.Delete(ctx, item.key)
+				storage.Put(ctx, append([]byte{containerKeyPrefix}, item.key...), item.value)
+			}
+
+			// Migrate owner-cid map.
+			if len(item.key) == 25 /* owner id size */ +containerIDSize {
+				storage.Delete(ctx, item.key)
+				storage.Put(ctx, append([]byte{ownerKeyPrefix}, item.key...), item.value)
+			}
+		}
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


container/container_contract.go
```diff
@@ -93,6 +95,32 @@ func _deploy(data interface{}, isUpdate bool) {
 	if isUpdate {
 		args := data.([]interface{})
 		common.CheckVersion(args[len(args)-1].(int))
+
```
roman-khimov, line 98: There has to be some check for the current contract version here, if it's `>= some` this migration is not needed.

container/container_contract.go
```diff
@@ -391,17 +419,24 @@ func Owner(containerID []byte) []byte {
...
-		if len(key) == containerIDSize {
-			count++
-		}
+		count++
 	}
 	return count
 }
 
+// Iterate iterates over all container IDs owned by the specified owner.
+func Iterate(owner []byte) iterator.Iterator {
```
roman-khimov, line 430: `ContainersOf`? Frankly, I'm thinking about some NEP-11 compatibility for this contract.

container/container_contract.go
```diff
@@ -391,17 +419,24 @@ func Owner(containerID []byte) []byte {
...
+	ctx := storage.GetReadOnlyContext()
+
+	if len(owner) == 0 {
+		return storage.Find(ctx, []byte(containerKeyPrefix), storage.KeysOnly|storage.RemovePrefix)
```
roman-khimov, line 434: Wow, this differs a lot from `ValuesOnly` below, this shouldn't be done, we want iterators to have consistent values. If you need some different behavior just add another method, it's not hard to do that. And makes this one only accept correct owners (like `tokensOf` does).

container/container_contract.go
```diff
@@ -62,6 +62,8 @@ const (
 
 	singleEstimatePrefix = "est"
 	estimateKeyPrefix    = "cnr"
+	containerKeyPrefix   = "cid"
+	ownerKeyPrefix       = "own"
```
roman-khimov, line 66: We have some 3-letter prefixes already unfortunately, but I'd use 1-byte for new ones anyway.

container/container_contract.go
```diff
@@ -93,6 +95,32 @@ func _deploy(data interface{}, isUpdate bool) {
...
+				value []byte
+			})
+			if len(item.key) == containerIDSize {
+				storage.Delete(ctx, item.key)
+				storage.Put(ctx, append([]byte(containerKeyPrefix), item.key...), item.value)
+			}
+		}
+
+		// Migrate owner-cid map.
+		it = storage.Find(ctx, []byte{}, storage.None)
```
roman-khimov, line 113: Why two iterations over the DB?

roman-khimov (changes requested):

container/container_contract.go
fyrchik: Our current policy is that we remove this code after release and update "min updating from" versions in `common` package. This way we can update contract multiple times, but do not have complex in `_deploy`.

container/container_contract.go
fyrchik: Fixed. Initially I wanted to use 'c' as a prefix which would conflict with `cnr`, but nobody cares really.

container/container_contract.go
fyrchik: Fixed.

container/container_contract.go
fyrchik: I mirrored `List` logic here, but we can iterate over `ownerKeyPrefix` too, fixed.

container/container_contract.go
fyrchik: Fixed.

container/container_contract.go
roman-khimov: Not very user-friendly, upgrading contracts multiple times in no fun especially given that some node versions work fine with some contract versions and don't work with others.

container/container_contract.go
```diff
@@ -700,29 +729,30 @@ func Version() int {
...
+	containerListKey = append(containerListKey, id...)
 	storage.Delete(ctx, containerListKey)
 
-	storage.Delete(ctx, id)
+	storage.Delete(ctx, append([]byte{containerKeyPrefix}, id...))
 }
 
 func getAllContainers(ctx storage.Context) [][]byte {
-	var list [][]byte
+	list := [][]byte{}
```
carpawell, line 749: what's the byte code diff with that lines?

tests/container_test.go
```diff
@@ -103,15 +103,56 @@ func TestContainerCount(t *testing.T) {
...
+
+		actual := make([][]byte, 0, len(expected))
+		for i := range arr {
+			id, ok := arr[i].Value().([]byte)
+			require.True(t, ok)
+			actual = append(actual, id)
+		}
+		require.ElementsMatch(t, expected, actual)
+	})
+	t.Run("check with `iterate`", func(t *testing.T) {
```
carpawell, line 139:
```suggestion
	t.Run("check with `containersOf`", func(t *testing.T) {
```

carpawell (approved):

container/container_contract.go
roman-khimov: Should be none, which means that the change is not needed.

alexvanin: I think empty `var list [][]byte` returns `[{"type":"Any"}]` on stack, while `list := [][]byte{}` returns `[{"type":"Array","value":[]}]`.

tests/container_test.go
fyrchik: Fixed

PR #306, opened 2023-01-27
Format Go code using go fmt

cthulhu-rider:


roman-khimov:
Can you also add a `make fmt` support? Something like this from NeoGo:
```
fmt:
        @gofmt -l -w -s $$(find . -type f -name '*.go'| grep -v "/vendor/")
```

roman-khimov (approved):

Issue #303, opened 2022-12-30
Drop notary-less code and data from all contracts except mainnet one


roman-khimov:
We can have notary subsystem on all public chains now and every private installation also uses it anyway, so notary-less code is only needed for the mainnet `neofs` contract. Some preliminary work has been done for this in the `enable-notary-in-public-chains` branch (https://github.com/nspcc-dev/neofs-contract/tree/enable-notary-in-public-chains), but it needs to be extended. The most important thing here is to ensure smooth migration to notary-enabled reality (including notary deposits and the `proxy` contract that is completely missing in non-notary setups).

roman-khimov:
* alphabet contract migration cost should add notary deposits for all IR and SNs
 * it also should send a half of its GAS to the proxy contract (assuming it's pre-deployed)
 * all contracts should check for their voting data to not contain any ongoing votes (panicing otherwise)
 * old voting data must be dropped
 * `notaryDisabled` settings must be dropped (except for `_deploy` parameters, it's cheaper to have them there for now to keep neofs-adm compatibility)

It's one-way migration. Contract are to be tested using real mainnet data (hi, #281).

PR #320, opened 2023-02-01
Document contracts' storage models

cthulhu-rider:
- [x] alphabet
- [x] audit
- [x] balance
- [x] container
- [x] neofs
- [x] neofsid
- [x] netmap
- [ ] nns
- [x] processing
- [x] proxy
- [x] reputation
- [x] subnet

### DIFF
--- a/audit/doc.go
+++ b/audit/doc.go
@@ -21,3 +21,19 @@ they make a list and get these AuditResultStructures from the audit contract.
 Audit contract does not produce notifications to process.
 */
 package audit
+
+/*
+Contract storage model.
+
+# Summary
+Key-value storage format:
+ - <epoch><cid><key> -> []byte
+   Data audit results encoded into NeoFS API binary protocol format. Results are
+   identified by triplet concatenation:
+    1. little-endian unsigned integer NeoFS epoch when audit was performed
+    2. 32-byte identifier of the NeoFS container under audit
+    3. 24-byte prefix of SHA-256 hash of the auditor's (Inner Ring) public key
+
+# Audit history
+Contracts stores results of the NeoFS data audits performed by the Inner Ring.
+*/
--- a/balance/doc.go
+++ b/balance/doc.go
@@ -77,3 +77,17 @@ when NeoFS contract has transferred GAS assets back to the user.
 	    type: Integer
 */
 package balance
+
+/*
+Contract storage model.
+
+# Summary
+Key-value storage format:
+ - 'MainnetGAS' -> int
+   total amount of Mainchain GAS deployed in the NeoFS network in Fixed12
+ - interop.Hash160 -> std.Serialize(Account)
+   balance sheet of all NeoFS users (here Account is a structure defined in current package)
+
+# Accounting
+Contract stores information about all NeoFS accounts.
+*/
--- a/container/doc.go
+++ b/container/doc.go
@@ -67,3 +67,52 @@ it in Container contract.
 	    type: Integer
 */
 package container
+
+/*
+Contract storage model.
+
+# Summary
+Current conventions:
+ <cid>: 32-byte container identifier (SHA-256 hashes of container data)
+ <owner>: 25-byte NEO3 account of owner of the particular container
+ <epoch>: little-endian unsigned integer NeoFS epoch
+
+Key-value storage format:
+ - 'netmapScriptHash' -> interop.Hash160
+   Netmap contract reference
+ - 'balanceScriptHash' -> interop.Hash160
+   Balance contract reference
+ - 'identityScriptHash' -> interop.Hash160
+   NeoFSID contract reference
+ - 'nnsScriptHash' -> interop.Hash160
+   NNS contract reference
+ - 'nnsRoot' -> interop.Hash160
+   NNS root domain zone for containers
+ - 'x<cid>' -> []byte
+   container descriptors encoded into NeoFS API binary protocol format
+ - 'o<owner><cid>' -> <cid>
+   user-by-user containers
+ - 'nnsHasAlias<cid>' -> string
+   domains registered for containers in the NNS
+ - 'cnr<epoch><cid>' + [10]byte -> std.Serialize(estimation)
+   estimation of the container size sent by the storage node. Key suffix is first
+   10 bytes of RIPEMD-160 hash of the storage node's public key
+   (interop.PublicKey). Here estimation is a type.
+ - 'est' + [20]byte -> []<epoch>
+   list of NeoFS epochs when particular storage node sent estimations. Suffix is
+   RIPEMD-160 hash of the storage node's public key (interop.PublicKey).
+
+# Setting
+To handle some events, the contract refers to other contracts.
+
+# Containers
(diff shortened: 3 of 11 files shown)


cthulhu-rider:
storage scheme eased after #303, I'll redo all progress and finish

cthulhu-rider:
we have WIP related to NNS (#266, #267), so it's better to postpone scheme description for this contract

balance/doc.go
```diff
@@ -77,3 +77,17 @@ when NeoFS contract has transferred GAS assets back to the user.
...
 */
 package balance
+
+/*
+Contract storage model.
+
+# Summary
+Key-value storage format:
+ - 'MainnetGAS' -> int
+   total amount of Mainchain GAS deployed in the NeoFS network
```
roman-khimov, line 87: In Fixed12? Looks like. And it's important since mainnet GAS is Fixed8.

audit/doc.go
```diff
@@ -21,3 +21,19 @@ they make a list and get these AuditResultStructures from the audit contract.
...
+
+/*
+Contract storage model.
+
+# Summary
+Key-value storage format:
+ - [64]byte -> []byte
+   Data audit results encoded into NeoFS API binary protocol format. Results are
+   identified by triplet concatenation:
+    1. 8-byte little-endian unsigned integer NeoFS epoch when audit was performed
```
roman-khimov, line 33: That's not correct, the length in fact is dynamic.

container/doc.go
```diff
@@ -67,3 +67,52 @@ it in Container contract.
...
 package container
+
+/*
+Contract storage model.
+
+# Summary
+Current conventions:
+ <cid>: 32-byte container identifier (SHA-256 hashes of container data)
+ <owner>: 25-byte NEO3 account of owner of the particular container
+ <epoch>: 8-byte little-endian unsigned integer NeoFS epoch
```
roman-khimov, line 78: Also a dynamic thing.

reputation/doc.go
```diff
@@ -16,3 +16,22 @@ Inner Ring nodes if data audit succeeds.
 Reputation contract does not produce notifications to process.
 */
 package reputation
+
+/*
+Contract storage model.
+
+Current conventions:
+ <peer>: binary unique identifier of the NeoFS Reputation system participant
+ <epoch>: 8-byte little-endian unsigned integer NeoFS epoch
```
roman-khimov, line 25: Not 8 bytes.

reputation/doc.go
```diff
@@ -16,3 +16,22 @@ Inner Ring nodes if data audit succeeds.
...
+
+# Summary
+Key-value storage format:
+ - 'c<peer><epoch>' -> int
```
roman-khimov, line 29: `<epoch><peer>`, same as below.

balance/doc.go
cthulhu-rider: how constant is this precision overall? Contract is NEP-17 so it provides `Decimals` method

maybe it's clearer to write `with precision determined by the current contract`? Exact precision is described in contract docs and, in theory, can be changed. In storage we have generic `int` anyway

audit/doc.go
cthulhu-rider: dropped length

reputation/doc.go
cthulhu-rider: fixed

balance/doc.go
roman-khimov: > how constant is this precision overall?

It can't change for a deployed NEP-17. So I doubt it can be changed in our contracts. Not until we change the internal token to something else.

> maybe it's clearer to write with precision determined by the current contract

The main problem is that we're talking main chain GAS here and it's known to have precision of 8, so one can expect 1 GAS stored as 100000000, while in fact that's not the case, it's an internal representation of this GAS which currently has precision of 12 (and as I've said _very_ unlikely to change).

cthulhu-rider: got ur point

clarified

roman-khimov (dismissed):

neofsid/doc.go
```diff
@@ -19,3 +19,18 @@ contract.
 NeoFSID contract does not produce notifications to process.
 */
 package neofsid
+
+/*
+Contract storage model.
```
notimetoname, line 24: what does that line mean?

netmap/doc.go
```diff
@@ -54,17 +54,10 @@ Key-value storage format:
    Container contract reference
  - 'balanceScriptHash' -> 20-byte script hash
    Balance contract reference
- - 'notary' -> bool
```
notimetoname, line 57: a separate commit?

audit/doc.go
```diff
@@ -21,3 +21,19 @@ they make a list and get these AuditResultStructures from the audit contract.
...
+
+/*
+Contract storage model.
+
+# Summary
+Key-value storage format:
+ - [64]byte -> []byte
+   Data audit results encoded into NeoFS API binary protocol format. Results are
+   identified by triplet concatenation:
+    1. little-endian unsigned integer NeoFS epoch when audit was performed
```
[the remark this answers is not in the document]
roman-khimov: It can be from 0 to 4.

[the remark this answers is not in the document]
notimetoname: yes, but is documented as `[64]byte` so i guess it is always 8 (as a reader, as i dev currently i have no idea cause haven't looked deep), also i found `// 24 + 32 (container ID length) + 8 (epoch length) = 64` comment. also, [i know](https://github.com/nspcc-dev/neofs-api-go/blob/a900856c556d0a23c983214b1b4058bca51e0590/audit/marshal.go#L42) that it is `FIxed64` field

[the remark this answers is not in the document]
roman-khimov: > [64]byte

I see what you mean now, yes, the size of the key is dynamic.

> little-endian unsigned integer

But that's still true.

neofsid/doc.go
cthulhu-rider: i don't know how to answer, it means itself 
u think its redundant/incorrect?

netmap/doc.go
cthulhu-rider: overall ur right, but here i thought that it's not really need, i wanna introduce single initial commit when all storage models are described and actual, and then change them with corresponding changes in the code

:+1: if it's worth separate commit anyway, i'll do

neofsid/doc.go
notimetoname: if it is a key for the whole file (`doc.go`) then sometimes the lines below are not `only` about the storage model; if it is a title for some paragraph then there are titles that are not the same as i see (every further titles are `# ...`)

roman-khimov (approved):

notimetoname (approved):

neofsid/doc.go
cthulhu-rider: the rest of the blocks are hints to the stored data, not a schema, but still part of the descriptive model. But I understand your position, it is also reasonable. If u dont mind let's leave it as it is, we can always change, it's cosmetic

AnnaShaleva (approved):

Issue #301, opened 2022-12-24
Improve contract update and migration


fyrchik:
Currently when updating from a very old version, we must update contracts multiple times, because we only support migration from a previous version. There were 2 reasons:
1. Avoid complex logic in contracts. We have only 1 "old" format and must migrate it to a single "new" format.
2. Easier testing (linear amount of user scenarios to test 1-2 + 2-3 + 3-4 vs possible 1-2,2-3,3-4 + 1-3,3-4 + 1-4 + 1-2,2-4 + something we will surely forget).

As @roman-khimov noticed this is not very user-friendly. One of the ways do fix this is to have  linear migration logic in the `_deploy` itself: we still migrate in a linear fashion, but can do it multiple times in a single update. This way we need to add a single migration when changing storage format, but also allow updating from old contract versions. Not to blow up contract size (neofs-adm would like to update contracts in a single transaction) we can also disallow updating from "very old" versions.

However, I would implement #281 first.

roman-khimov:
I doubt contract size is much of a problem for us, we can have this logic in `_deploy` for quite some time. Eventually (depending on support cycles mostly) of course old transition code can be dropped with notices that version X.Y.Z can be upgraded to only from A.B.C or newer.

roman-khimov:
> neofs-adm would like to update contracts in a single transaction

It does something like three of them already and this can be changed if needed, I think.

roman-khimov:
#324 has implemented the new convention, effectively.

PR #324, opened 2023-02-10
Stop supporting non-notary deployments

### LINKED ISSUE #303: Drop notary-less code and data from all contracts except mainnet one

cthulhu-rider:
* closes #303

- [ ] clean storage
- [ ] remove non-notary node from methods
- [ ] distribute GAS on Alphabet contract update


audit/audit_contract.go
```diff
@@ -36,39 +36,50 @@ func (a auditHeader) ID() []byte {
...
+
+	notaryVal := storage.Get(ctx, notaryDisabledKey)
+	if notaryVal == nil {
+		runtime.Log("contract is already notarized")
+		return
 	}
 
-	runtime.Log("audit contract initialized")
+	storage.Delete(ctx, notaryDisabledKey)
+	storage.Delete(ctx, "netmapScriptHash")
```
roman-khimov, line 78: Safe to do on notary-enabled networks?

neofsid/neofsid_contract.go
```diff
@@ -35,30 +33,57 @@ func _deploy(data interface{}, isUpdate bool) {
...
+		switchToNotary(ctx, args)
+
 		return
 	}
 
 	args := data.(struct {
 		notaryDisabled bool
 		addrNetmap     interop.Hash160
-		addrContainer  interop.Hash160
+		_              interop.Hash160 // Netmap contract from non-notary legacy
```
roman-khimov, line 45: Isn't it a `container` contract address?

netmap/netmap_contract.go
```diff
@@ -200,73 +219,27 @@ func Update(script []byte, manifest []byte, data interface{}) {
...
+// deployments only.
 func InnerRingList() []common.IRNode {
-	ctx := storage.GetReadOnlyContext()
-	pubs := getIRNodes(ctx)
-	nodes := []common.IRNode{}
-	for i := range pubs {
-		nodes = append(nodes, common.IRNode{PublicKey: pubs[i]})
-	}
-	return nodes
+	// TODO: consider method removal
```
roman-khimov, line 225: Can be removed if we're sure notary-enabled node never calls them.

netmap/netmap_contract.go
```diff
@@ -278,78 +251,37 @@ func AddPeerIR(nodeInfo []byte) {
...
-
-		n := common.Vote(ctx, id, nodeKey)
-		if n < threshold {
-			return
-		}
-
-		common.RemoveVotes(ctx, id)
+	common.CheckWitness(publicKey)
+
+	// TODO: is it good approach? We could always call AddPeerIR by the Alphabet,
```
roman-khimov, line 273: This method shouldn't be called in notary-enabled env anyway (except for storage nodes, but it's a no-op then?), so maybe it can be stubbed out.

subnet/subnet_contract.go
```diff
@@ -61,15 +60,43 @@ func _deploy(data interface{}, isUpdate bool) {
...
+// switchToNotary removes value stored by 'notary' key.
+func switchToNotary(ctx storage.Context, args []interface{}) {
+	const notaryDisabledKey = "notary" // non-notary legacy
+
+	notaryVal := storage.Get(ctx, notaryDisabledKey)
+	if notaryVal == nil {
+		runtime.Log("contract is already notarized")
+		return
+	}
+
```
roman-khimov, line 94: BTW, this contract can have voting data as well.

reputation/reputation_contract.go
```diff
@@ -24,23 +23,47 @@ func _deploy(data interface{}, isUpdate bool) {
...
+	const notaryDisabledKey = "notary" // non-notary legacy
+
+	notaryVal := storage.Get(ctx, notaryDisabledKey)
+	if notaryVal == nil {
+		runtime.Log("contract is already notarized")
+		return
+	}
+
+	storage.Delete(ctx, notaryDisabledKey)
+
```
roman-khimov, line 61: And this one too.

alphabet/alphabet_contract.go
```diff
@@ -50,6 +51,10 @@ func _deploy(data interface{}, isUpdate bool) {
 		total          int
 	})
 
+	if args.notaryDisabled {
```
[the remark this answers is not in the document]
cthulhu-rider: You mean fail only if the original mode was notary, and if non-notary - just do nothing? I don’t like the unconditional transition to a notary mode, because update "request" can explicitly state the requirement "make it non-notary".

alphabet/alphabet_contract.go
```diff
@@ -38,6 +36,9 @@ func _deploy(data interface{}, isUpdate bool) {
 	if isUpdate {
 		args := data.([]interface{})
 		common.CheckVersion(args[len(args)-1].(int))
+
+		switchToNotary(ctx, args)
```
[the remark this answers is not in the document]
cthulhu-rider: Wouldn't the check above be enough for us? It requires version to grow only.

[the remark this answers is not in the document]
roman-khimov: #301, right now we upgrade from 0.16.0+ to 0.17.0, but when someone will be upgrading 0.17.0+ to any later version this code is not needed (even though `CheckVersion` will allow 0.16.0).

alphabet/alphabet_contract.go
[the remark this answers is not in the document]
roman-khimov: This flag means nothing after the transition. `make is non-notary` is not a valid thing to request from 0.17.0+ contracts, we move to notary-only and that's it.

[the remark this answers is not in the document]
cthulhu-rider: > is not a valid thing

So we panic in this case, isn't that correct? I guess I missed your point.

alphabet/alphabet_contract.go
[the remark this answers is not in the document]
cthulhu-rider: Added condition for v0.17.0. I suggest to do refactoring with step-by-step approach separately within #301.

audit/audit_contract.go
cthulhu-rider: I guess, these are only usages:
https://github.com/nspcc-dev/neofs-contract/blob/5b007f429fa980ccd3879e328010aa771692e320/audit/audit_contract.go#L98-L100
https://github.com/nspcc-dev/neofs-contract/blob/5b007f429fa980ccd3879e328010aa771692e320/audit/audit_contract.go#L181-L202

netmap/netmap_contract.go
cthulhu-rider: Method docs require this. We'll definitely double-check that in `neofs-node`. Anyway, if some notary-adopted code calls the method, it'll encounter `unknown method` exception.

netmap/netmap_contract.go
cthulhu-rider: The method requires transaction to be signed by the candidate itself (see `common.CheckWitness` above). So, we can completely abandon the method only by checking on the Inne Ring side. What do you think?

subnet/subnet_contract.go
cthulhu-rider: How did you determine it? I don't see the corresponding code. For certainty?

roman-khimov: It uses `common.Vote()` in `Put()`

alphabet/alphabet_contract.go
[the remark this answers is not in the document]
roman-khimov: We can ignore this value (not use in any way) before we remove it completely (but this will take some time). Although panic may help us detect invalid arguments (`notaryDisabled == true`).

alphabet/alphabet_contract.go
```diff
@@ -60,14 +65,52 @@ func _deploy(data interface{}, isUpdate bool) {
...
+	if notaryVal == nil {
+		runtime.Log(contractName + " contract is already notarized")
+		return
+	} else if notaryVal.(bool) {
+		proxyContract := args[2].(interop.Hash160)
+		if len(proxyContract) != interop.Hash160Len {
+			panic("address of the Proxy contract is missing or invalid")
+		}
+
+		// TODO: should we check if Proxy contract is already deployed?
```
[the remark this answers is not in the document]
cthulhu-rider: Then we need to accept NNS address as update parameter, am I right? I don't see any alternatives now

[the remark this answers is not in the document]
cthulhu-rider: Check out https://github.com/nspcc-dev/neofs-contract/pull/324/commits/665433990f269286e76e51c0560cfc66b19b0688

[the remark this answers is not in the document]
roman-khimov: It's number one, can be retrieved via https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/interop@v0.0.0-20230208100456-1d6e48ee78e5/native/management#GetContractByID.

balance/balance_contract.go
```diff
@@ -63,31 +59,64 @@ func _deploy(data interface{}, isUpdate bool) {
...
 	args := data.(struct {
 		notaryDisabled bool
-		addrNetmap     interop.Hash160
-		addrContainer  interop.Hash160
+		_, _           interop.Hash160 // Netmap and Container contract from non-notary legacy
 	})
 
-	if len(args.addrNetmap) != interop.Hash160Len || len(args.addrContainer) != interop.Hash160Len {
-		panic("incorrect length of contract script hash")
+	if args.notaryDisabled {
```
roman-khimov, line 86: If we no longer need any parameters for a contract, let's drop this code completely (and don't touch `data` at all). This applied to any contract that needs no parameters (others have to keep the same structure for now, so checking `notaryDisabled` there makes sense).

roman-khimov (commented):
Adding these now, will check tests later.

alphabet/alphabet_contract.go
```diff
@@ -60,14 +78,115 @@ func _deploy(data interface{}, isUpdate bool) {
...
+			panic("wrong address of the Proxy contract")
+		}
+
+		if !common.TryPurgeVotes(ctx) {
+			panic("pending vote detected")
+		}
+
+		// distribute GAS:
+		//  - 50% to Proxy contract
+		//  - the rest is evenly distributed between Inner Ring and storage nodes
```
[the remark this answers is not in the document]
cthulhu-rider: How much? Maybe a quarter?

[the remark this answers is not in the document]
roman-khimov: Seems to be OK. 75% of the balance should be sufficient for nodes to operate (and proxy to have enough GAS for notary-enabled things).

tests/migration/model.go
```diff
@@ -0,0 +1,137 @@
...
+// ItemGroupModel describes migration model of some storage item group of the
+// fixed Neo contract.
+type ItemGroupModel struct {
+	keyPattern binPattern
+	modifier   func(prev kv) kv
+}
+
+// UpdateModel describes storage migration model of some Neo contract. Zero update model
+// describes empty post-update contract storage.
+type UpdateModel struct {
```
roman-khimov, line 17: This model describes change set in terms of KV pairs, which is fine when:
 * we delete some key or a set of keys with some prefix and want to ensure it's gone
 * we specifically add some key or a set of keys and want to ensure they're present
 * we want to make sure no excessive elements are added

But it's suboptimal when we're dealing with changes. Consider our netmap 0.16.0 migration problem (df2bad0f2fcddc1d405be3fa8003eec3c8d1d1b1), if we have this omission in the contract we're quite likely to have the same omission in the test, but calling contract APIs could quickly reveal the problem, so I'd say we should describe these in terms of API calls (test invocations) that should have the same (or intentionally different if needed) result before the migration and after the migration.

tests/migration/storage.go
```diff
@@ -0,0 +1,152 @@
...
+//     blockchain environment
+//  1. compile and deploy NeoFS NNS contract from sources located in '../nns' directory
+//  2. fetch tested contract executable by contractName parameter
+//  3. deploy tested contract in the test blockchain
+//  4. compile tested contract from sources located in '../contractName' directory
+//  5. update the tested contract with executables compiled in step 4
+//  6. check that test contract's storage is empty
+//
+// See Options to tune the default execution.
+func TestRemoteStorageMigration(tb testing.TB, blockChainRPCEndpoint, contractName string, opts Options) {
```
roman-khimov, line 71: As already discussed the "remoteness" of this procedure is better be handled by a script downloading data from this external source. Actual tests should operate with static data set stored somewhere in `testdata`.

tests/migration/storage.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// 4.
+	if opts.SourceCodeDir == "" {
+		opts.SourceCodeDir = SameLevelFolder(contractName)
+	}
+
+	// 5.
+	contractUpd := newContractUpdater(tb, bLocal, opts.SourceCodeDir, localContract)
+
+	for _, updFail := range opts.UpdateFailures {
```
roman-khimov, line 121: These better be independent `t.Run()` and maybe better in the test itself.

tests/migration/storage.go
```diff
@@ -0,0 +1,152 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/nspcc-dev/neo-go/pkg/util"
+	"github.com/stretchr/testify/require"
+)
+
+// Options groups optional parameters for TestRemoteStorageMigration.
```
roman-khimov, line 11: It tries to cover every possible case, but probably we can simplify this. The way I'd like to see these tests is:
 * get previously stored data, init storage with it
 * for every test create a secondary memcache (to keep initial one clean between tests)
 * initialize Blockchain with it
 * perform action
 * check result
 * drop this instance and move to the new test

Current scheme limits `Options` to a single successful run, but we might as well have a number of them (even if they're based on the same data, it can be tweaked in some ways by test).

roman-khimov (commented):
Have you tried it on the mainnet data?

tests/migration/model.go
cthulhu-rider: IMO these are two different tests:
1. migration of storage items from one format to another
2. ability of new contract executable to correctly decode new format

2nd won't help us if there is no corresponding values in the contract at all. But I agree that such API extra check will help us to feel more comfortable.

netmap/netmap_contract.go
cthulhu-rider: @roman-khimov can we remove this condition?

roman-khimov: No, it'll break things, we have to keep it so that the SN could successfully test-run the method and IR could successfully see the event from a test run of its own, but then ignore it. I will just link nspcc-dev/neofs-node#2278 here.

roman-khimov (approved):

PR #333, opened 2023-04-13
common: ignore stale votes in TryPurgeVotes

roman-khimov:
On a real healthy network we _always_ have something there. Consider seven IRs doing something, we only need five signatures, so they're added and then dropped when we have enough of them, but excessive transactions from the other IRs still come and they will be added again into the same voting structure.

But according to the definition of blockDiff we can ignore stale votes safely (they're not counted for votes anyway).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,19 @@
 # Changelog
 Changelog for NeoFS Contract
 
+## [Unreleased]
+
+### Added
+
+### Updated
+
+### Removed
+
+### Fixed
+- migration of non-notary network to notarized one with stale votes
+
+### Updating from v0.17.0
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,19 @@
 # Changelog
 Changelog for NeoFS Contract
 
+## [Unreleased]
```
cthulhu-rider, line 4: lets add diff link for this

0.17.0 link is also missing btw

cthulhu-rider (approved):

CHANGELOG.md
roman-khimov: Yeah, I've noticed. Though to create another PR, but hey, let's have it here.

cthulhu-rider (approved):

PR #341, opened 2023-05-03
Drop old events

roman-khimov:
Clean up a bit.

### DIFF
--- a/reputation/reputation_contract.go
+++ b/reputation/reputation_contract.go
@@ -85,7 +85,7 @@ func Update(script []byte, manifest []byte, data interface{}) {
 }
 
 // Put method saves DataAuditResult in contract storage. It can be invoked only by
-// Inner Ring nodes. It does not require multisignature invocations.
+// storage nodes with Alphabet assistance (multisignature witness).
 //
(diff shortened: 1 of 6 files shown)


reputation/reputation_contract.go
```diff
@@ -94,12 +94,7 @@ func Put(epoch int, peerID []byte, value []byte) {
 	ctx := storage.GetContext()
 
 	multiaddr := common.AlphabetAddress()
-	alphabetCall := runtime.CheckWitness(multiaddr)
-
-	if !alphabetCall {
-		runtime.Notify("reputationPut", epoch, peerID, value)
-		return
-	}
+	common.CheckAlphabetWitness(multiaddr)
```
AnnaShaleva, line 97: BTW, the behaviour has been changed, is it OK? The previous behaviour was to return from the function if not witnessed by alphabet, and the current one is to panic.

roman-khimov: Yes, this is explained in the commit message.

AnnaShaleva: Oh, sorry, I've checked the "Files changed" section only.

AnnaShaleva (approved):

Issue #334, opened 2023-04-17
Extend NNS contract's _deploy with predefined TLDs


roman-khimov:
Accept an array of `{name, email}` params and register TLD with some default refresh/retry/expire/ttl values. TLD should be owned by the committee.

Related to nspcc-dev/neofs-node#2195. See also neo-project/non-native-contracts#32 (although it has a domain set in the code, while we want them to be in `_deploy` parameters).

roman-khimov:
Deploy only.

cthulhu-rider:
> TLD should be owned by the committee

do we require committee witness to work with these domains? Or each committee member becomes a solo admin?

During NeoFS initialization NNS contract is deployed with `bootstrap` domain owned by the committee. Later `az` member attempts to register `designate-committee-notary-tx.boostrap` domain but if we require committee witness then `az` should provide committee multi-signature.

roman-khimov:
IIUC we currently have somewhat broken TLDs in that they require a TLD owner signature to register a domain. Which btw means that we can't change committee keys (a particular account for particular set of keys is set in stone as an owner). Compare that to C# NNS root entry management --- only committee (and not a specific address) can register and manage it, but then anyone can register a domain using any of the available roots.

We need something like that, TLD managed by committee (not having a real owner), but then `Register` allowing to grab any domain (and only doing admin checks for third-level domains). This should solve the problem.

CC @AnnaShaleva.

AnnaShaleva:
> they require a TLD owner signature to register a domain

That's not exactly true. They require TLD *admin's* signature to register the TLD's children domain. And admin can always be changed by owner.

>  Which btw means that we can't change committee keys
Thus, we actually can change the committe keys, but we should probably change the TLD domain admin as far.

> Compare that to C# NNS root entry management --- only committee (and not a specific address) can register and manage it
That's true, only committee can register TLD. And after that anyone can register subdomains for the existing TLD, and committe signature isn't required for that.

PR #344, opened 2023-05-28
nns: Register pre-defined TLDs for the committee on deployment stage

cthulhu-rider:


roman-khimov (commented):
The concept of "committee" domain is excessive. We need _all_ TLDs to be managed by committee. And then some TLD could be created in `_deploy`. And then _anyone_ should be able to create regular domains using already existing TLDs (no need to make committee members admins). Like in #334.

cthulhu-rider:
lets then explicitly denote the layer of top-level domains controlled by the committee. 2nd level domains can be freely registered by anybody. Any deeper level requires owner or admin authorization.

cthulhu-rider:
> The concept of "committee" domain is excessive

agree overall, but what behavior do u expect from methods of ownership like https://pkg.go.dev/github.com/nspcc-dev/neofs-contract/nns#OwnerOf called with TLD?

cthulhu-rider:
@roman-khimov @AnnaShaleva i've updated changes

do u think we need dedicated `registerTLD` method? Let me remind u that now `register` method takes the domain owner as a parameter, which from now will ignored during TLD registration.

i'm ok to leave `register` as it is. With `registerTLD`, it'd make sense to ban TLD registrations via `register`

roman-khimov:
> OwnerOf called with TLD

It's just not a proper domain, so it should behave as if this domain was never registered. I think that's similar to C# NNS behavior for TLDs.

cthulhu-rider:
> should behave as if this domain was never registered

so smth like
```
OwnerOf("bob.com") -> Bob
OwnerOf("com") -> panic("domain not exists")
```
?

or it'd be better to panic with `TLD not allowed`?

roman-khimov:
IIUC `inexistent.com` will fail with `panic("token not found")`, that's OK for TLD too.

nns/namestate.go
```diff
@@ -22,6 +23,10 @@ func (n NameState) ensureNotExpired() {
 
 // checkAdmin panics if script container is not signed by the domain name admin.
 func (n NameState) checkAdmin() {
+	if n.Owner == nil {
+		checkCommittee()
+		return
+	}
```
AnnaShaleva, line 29: If we don't check (and don't allow) `Admin` for TLDs, then we probably should explicitly fail `setAdmin` for it. There's a `common.CheckOwnerWitness(ns.Owner)` that will fail for an empty owner, but we should probably give more detailed error in this case.

nns/nns_contract.go
```diff
@@ -276,27 +282,37 @@ func Register(name string, owner interop.Hash160, email string, refresh, retry,
...
-		}
-		if parentExpired(ctx, 1, fragments) {
-			panic("one of the parent domains is not registered")
-		}
-		parentKey := getTokenKey([]byte(name[len(fragments[0])+1:]))
+		putNameStateWithKey(ctx, getTokenKey([]byte(name)), NameState{
+			Name: name,
+			// NNS expiration is in milliseconds
+			Expiration: int64(runtime.GetTime() + expire*1000),
+		})
```
AnnaShaleva: `ownerOf` method of such domain will return an empty hash. Is it correct from the user's side of view?

nns/nns_contract.go
```diff
@@ -276,27 +282,37 @@ func Register(name string, owner interop.Hash160, email string, refresh, retry,
...
+		putSoaRecord(ctx, name, email, refresh, retry, expire, ttl)
+		return true
```
AnnaShaleva, line 300: No notification will be emitted on registering TLDs which is correct, because it's not a token anymore. However, it still may be desired thing to see from the user side. Don't we want to notify about TLDs registering (probably, with some special notification) as far?

nns/nns_contract.go
```diff
@@ -336,17 +345,35 @@ func Register(name string, owner interop.Hash160, email string, refresh, retry,
...
 		updateTotalSupply(ctx, +1)
 	}
-	ns := NameState{
+	saveDomain(ctx, name, email, refresh, retry, expire, ttl, owner)
+	updateBalance(ctx, []byte(name), owner, +1)
+	postTransfer(oldOwner, owner, []byte(name), nil)
+	return true
+}
+
+//nolint:unused
```
AnnaShaleva, line 354: Remove this comment, add a comment about what this method does.

nns/nns_contract.go
```diff
@@ -336,17 +345,35 @@ func Register(name string, owner interop.Hash160, email string, refresh, retry,
...
+		refresh = 3600
+		retry   = 600
+		expire  = 10 * 365 * 24 * 60 * 60 // 10 years
+		ttl     = 3600
+	)
+	var committeeOwner interop.Hash160
+	saveDomain(ctx, name, email, refresh, retry, expire, ttl, committeeOwner)
+}
+
+func saveDomain(ctx storage.Context, name, email string, refresh, retry, expire, ttl int, owner interop.Hash160) {
```
AnnaShaleva, line 369: Ditto here and below, we'd better keep all the methods documented, NNS is being changed too often.

nns/nns_contract.go
```diff
@@ -273,21 +287,16 @@ func Register(name string, owner interop.Hash160, email string, refresh, retry,
...
 		}
-		storage.Put(ctx, tldKey, 0)
-		putNameStateWithKey(ctx, getTokenKey([]byte(name)), NameState{
-			Name: name,
-			// NNS expiration is in milliseconds
-			Expiration: int64(runtime.GetTime() + expire*1000),
-		})
-		putSoaRecord(ctx, name, email, refresh, retry, expire, ttl)
+		putTLD(ctx, tldKey)
+		saveDomain(ctx, name, email, refresh, retry, expire, ttl, nil)
```
AnnaShaleva, line 299: Is it OK to have `nil` instead of an empty `interop.Hash160`?

cthulhu-rider:
after discussion with @roman-khimov we have approved the behavior of TLD as a non-existent token for which operations like `ownerOf` work like with missing tokens. We also settled on the need to provide a dedicated `registerTLD` method and make `register` to work with "true" tokens, i.e. domains L2+.

@AnnaShaleva ty for the review, some of the discussions will be resolved after update I think

nns/nns_contract.go
cthulhu-rider: removed, added

nns/namestate.go
cthulhu-rider: fixed

nns/nns_contract.go
cthulhu-rider: fixed

nns/nns_contract.go
cthulhu-rider: currently in NeoFS nobody listens to these notifications. I don't see any practical need right now, but maybe u or @roman-khimov does?

nns/nns_contract.go
cthulhu-rider: tried to clarify what's going on

nns/nns_contract.go
cthulhu-rider: yes, `checkAdmin` checks `len(NameState.Owner) == 0` so IIUC there is no difference b/w `nil` and `{}`

nns/nns_contract.go
roman-khimov: It's a good style in general to have events for significant state changes. But this particular case is not handled in the C# contract as well, so if we don't know how to use it, we can omit it for now.

roman-khimov (commented):
I think we need some `listTLDs` iterator to be able to retrieve them if regular `tokens` don't work.

nns/nns_contract.go
```diff
@@ -374,16 +375,31 @@ func Register(name string, owner interop.Hash160, email string, refresh, retry,
...
+// RegisterTLD panics with 'TLD already exists' if domain already exists.
+func RegisterTLD(name, email string, refresh, retry, expire, ttl int) {
+	checkCommittee()
+	saveCommitteeDomain(storage.GetContext(), name, email, refresh, retry, expire, ttl)
+}
+
+// saveCommitteeDomain marks TLD as registered via prefixRoot<name> storage
+// record and saves domain state calling saveDomain with given parameters and
+// empty owner. The name MUST be a valid TLD name.
+func saveCommitteeDomain(ctx storage.Context, name, email string, refresh, retry, expire, ttl int) {
```
AnnaShaleva, line 390: This method isn't used anywhere else, thus please, move this code directly into RegisterTLD. NNS has a lot of tiny methods-helpers, and it's already hard to distinguish them, don't make it harder.

nns/nns_contract.go
```diff
@@ -137,15 +137,29 @@ func TotalSupply() int {
...
-// OwnerOf returns the owner of the specified domain.
+// OwnerOf returns the owner of the specified domain. The tokenID domain MUST
+// NOT be a TLD.
 func OwnerOf(tokenID []byte) interop.Hash160 {
+	// TODO: same done in getNameState, don't do twice
+	fragments := std.StringSplit(string(tokenID), ".")
+	if len(fragments) == 1 {
+		panic("token not found")
+	}
+
```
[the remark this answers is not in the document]
cthulhu-rider: yes we can, but why? ownership property is not defined for them as it is not a token and ownership is not transferred

@roman-khimov

nns/nns_contract.go
```diff
@@ -449,11 +487,19 @@ func checkBaseRecords(typ RecordType, data string) bool {
...
 	tokenID := []byte(tokenIDFromName(name))
 	if !checkBaseRecords(typ, data) {
 		panic("invalid record data")
 	}
+
+	// TODO: same done in getNameState, don't do twice
+	fragments := std.StringSplit(string(tokenID), ".")
+	if len(fragments) == 1 {
+		panic("token not found")
+	}
```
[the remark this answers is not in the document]
cthulhu-rider: i guess no, they are like abstract roots and all data is recorded starting from L2 domains, am i wrong? @roman-khimov

nns/nns_contract.go
cthulhu-rider: it's called in `RegisterTLD` and `_deploy`, just wanted to share similar instructions. But i don't mind to unwrap it, do u think it's worth it?

cthulhu-rider:
> I think we need some `listTLDs` iterator

btw contract provides `roots`, this is what you need? I've added test 92aa3edf9ce96fca4b6b06ed974025678c8b5a37

nns/nns_contract.go
```diff
@@ -183,10 +197,18 @@ func TokensOf(owner interop.Hash160) iterator.Iterator {
 }
 
 // Transfer transfers the domain with the specified name to a new owner.
+// The tokenID MUST NOT be a TLD.
 func Transfer(to interop.Hash160, tokenID []byte, data interface{}) bool {
 	if !isValid(to) {
 		panic(`invalid receiver`)
 	}
+
+	// TODO: same done in getNameState, don't do twice
```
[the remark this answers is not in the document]
cthulhu-rider: will do after core changes will be approved, this is cosmetic

nns/nns_contract.go
```diff
@@ -336,17 +345,39 @@ func Register(name string, owner interop.Hash160, email string, refresh, retry,
...
+}
+
+func saveCommitteeDomain(ctx storage.Context, name, email string) {
+	putTLD(ctx, makeTLDKey(name))
+
+	// TODO: values from NeoFS ADM, check
+	const (
+		refresh = 3600
+		retry   = 600
+		expire  = 10 * 365 * 24 * 60 * 60 // 10 years
```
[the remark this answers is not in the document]
cthulhu-rider: dropped todo

nns/nns_contract.go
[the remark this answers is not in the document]
roman-khimov: Yes, this just makes no sense for TLD.

nns/nns_contract.go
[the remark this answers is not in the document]
roman-khimov: Yes, there is no need to add any records to "container" or "neofs" and they're not tokens, so this restriction is OK.

cthulhu-rider:
> Why is this still a draft?

waited for meaningful discussions to be resolved. Now it seems to be finished, i'll do https://github.com/nspcc-dev/neofs-contract/pull/344#discussion_r1229252443 and open

nns/nns_contract.go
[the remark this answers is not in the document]
cthulhu-rider: done 613475b4c20e4842c277eb792e38ec0da47e3163

nns/nns_contract.go
```diff
@@ -260,6 +260,12 @@ func parentExpired(ctx storage.Context, first int, fragments []string) bool {
 }
 
 // Register registers a new domain with the specified owner and name if it's available.
+//
+// Access rules::
+//   - TLD can be registered only by the committee
+//   - 2nd-level domain can be registered by anyone
+//   - starting from the 3rd level, the domain can only be registered by the
+//     owner or administrator (if any) of the previous level domain
```
AnnaShaleva, line 309: BTW, current implementation doesn't support more than 3 levels.

nns/nns_contract.go
```diff
@@ -374,16 +374,31 @@ func Register(name string, owner interop.Hash160, email string, refresh, retry,
 	return true
 }
 
-func saveCommitteeDomain(ctx storage.Context, name, email string) {
-	putTLD(ctx, makeTLDKey(name))
+// RegisterTLD registers new top-level domain. RegisterTLD MUST be called by the
+// committee only. Name MUST be a valid TLD.
+//
+// RegisterTLD panics with 'TLD already exists' if domain already exists.
+func RegisterTLD(name, email string, refresh, retry, expire, ttl int) {
```
AnnaShaleva, line 378: Do we have an issue for neofs-adm to adapt these changes?

nns/nns_contract.go
cthulhu-rider: didn't pay attention..
is this an intentional/desired limitation? What does C# say?

roman-khimov: I doubt it's intentional, do we have an issue for it?

cthulhu-rider: @AnnaShaleva where exactly is this limitation? I didn’t find it from the sources, the test with L4 also passes

AnnaShaleva (approved):

nns/nns_contract.go
cthulhu-rider: found that this has already been changed f12c44fc786a9ed4e6ff24a38be3cde7cd624ee9

roman-khimov (approved):

PR #345, opened 2023-06-05
Makefile: generate RPC bindings

AnnaShaleva:
~~Waiting for 0.102.0 release of neo-go to convert from draft.~~

UPD: we'll use fresh NeoGo master to build SCs and RPC bindings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.avm
 *.nef
 *.manifest.json
+bindings_config.yml
 config.json
 /vendor/
 .idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 *.avm
 *.nef
-*.manifest.json
 bindings_config.yml
 config.json
 /vendor/
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ export GOBIN ?= $(shell pwd)/bin
 export CGO_ENABLED=0
 NEOGO ?= $(GOBIN)/cli
 VERSION ?= $(shell git describe --tags --dirty --match "v*" --always --abbrev=8 2>/dev/null || cat VERSION 2>/dev/null || echo "develop")
-NEOGOORIGMOD = github.com/nspcc-dev/neo-go
+NEOGOORIGMOD = github.com/nspcc-dev/neo-go@v0.101.5-0.20230802075307-afa4530c7d73
 NEOGOMOD = $(shell go list -f '{{.Path}}' -m $(NEOGOORIGMOD))
 NEOGOVER = $(shell go list -f '{{.Version}}' -m $(NEOGOORIGMOD) | tr -d v)
 
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,9 @@ nns_sc = nns
 
 define sc_template
 $(2)$(1)/$(1)_contract.nef: $(2)$(1)/$(1)_contract.go
-	$(NEOGO) contract compile -i $(2)$(1) -c $(if $(2),$(2),$(1)/)config.yml -m $(2)$(1)/config.json -o $(2)$(1)/$(1)_contract.nef
+	$(NEOGO) contract compile -i $(2)$(1) -c $(if $(2),$(2),$(1)/)config.yml -m $(2)$(1)/config.json -o $(2)$(1)/$(1)_contract.nef --bindings $(2)$(1)/bindings_config.yml
+	mkdir -p rpc/$(1)
+	$(NEOGO) contract generate-rpcwrapper -o rpc/$(1)/rpcbinding.go -m $(2)$(1)/config.json --config $(2)$(1)/bindings_config.yml
 
 $(if $(2),$(2)$(1)/$(1)_contract.go: alphabet/alphabet.go alphabet/alphabet.tpl
 	go run alphabet/alphabet.go
@@ -57,6 +59,8 @@ test:
 clean:
 	find . -name '*.nef' -exec rm -rf {} \;
 	find . -name 'config.json' -exec rm -rf {} \;
+	find . -name 'bindings_config.yml' -exec rm -rf {} \;
+	find . -name 'rpc' -type d -prune -exec rm -rf {} \;
 	rm -rf ./bin/
 
 mr_proper: clean
--- a/README.md
+++ b/README.md
@@ -37,28 +37,21 @@ Sidechain contracts:
 
 To compile smart contracts you need:
 
--   [neo-go](https://github.com/nspcc-dev/neo-go) >= 0.99.2
+-   [neo-go](https://github.com/nspcc-dev/neo-go) >= 0.102.0
 
 ## Compilation
 
 To build and compile smart contract, run `make all` command. Compiled contracts
 `*_contract.nef` and manifest `config.json` files are placed in the 
-corresponding directories. 
+corresponding directories. Generated RPC binding files `rpcbinding.go` are
+placed in the corresponding `rpc` directories.
 
 ```
 $ make all
-/home/user/go/bin/cli contract compile -i alphabet -c alphabet/config.yml -m alphabet/config.json -o alphabet/alphabet_contract.nef
-/home/user/go/bin/cli contract compile -i audit -c audit/config.yml -m audit/config.json -o audit/audit_contract.nef
-/home/user/go/bin/cli contract compile -i balance -c balance/config.yml -m balance/config.json -o balance/balance_contract.nef
-/home/user/go/bin/cli contract compile -i container -c container/config.yml -m container/config.json -o container/container_contract.nef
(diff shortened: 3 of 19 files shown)


AnnaShaleva:
Also need to bump minimum required go version up to 1.18.

Makefile
```diff
@@ -31,7 +31,9 @@ nns_sc = nns
 
 define sc_template
 $(2)$(1)/$(1)_contract.nef: $(2)$(1)/$(1)_contract.go
-	$(NEOGO) contract compile -i $(2)$(1) -c $(if $(2),$(2),$(1)/)config.yml -m $(2)$(1)/config.json -o $(2)$(1)/$(1)_contract.nef
+	$(NEOGO) contract compile -i $(2)$(1) -c $(if $(2),$(2),$(1)/)config.yml -m $(2)$(1)/config.json -o $(2)$(1)/$(1)_contract.nef --guess-eventtypes --bindings $(2)$(1)/bindings_config.yml
```
roman-khimov, line 34: `--guess-eventtypes`?

AnnaShaleva: Yes, and I'm strongly for the usage of `--guessed-eventtypes` flag, because it will properly handle all complicated event types without additional configuration. Although events are pretty simple in these contracts.

roman-khimov: To me this implies not having event in configuration files. But we have them now. And removing them is a bit scary to me.

roman-khimov: The only complex case is probably
```
      - name: keys
        type: Array
```
But I think it should be handled fine without guesses as well.

AnnaShaleva: It will, but without guesses its type will be `[]any`, because it's impossible to restore the type of array elements from the config without additional config modification:
```
// BindEvent represents "Bind" event emitted by the contract.
type BindEvent struct {
	User []byte
	Keys []any
}
```
And with guesses it has the proper and nice type:
```
// BindEvent represents "Bind" event emitted by the contract.
type BindEvent struct {
	User []byte
	Keys keys.PublicKeys
}
```

roman-khimov: What's gonna happen if I'm to write
```
      - name: keys
        type: Map
```
and compile with `--guess-eventtypes`?

AnnaShaleva: Currently the result with this flag will contain real type of the event parameters, and we have https://github.com/nspcc-dev/neo-go/issues/3027 to be resolved:
```
							// TODO: see the TestCompile_GuessEventTypes, "SC parameter type mismatch" section,
							// do we want to compare with actual.RealType? The conversion code is emitted by the
							// compiler for it, so we expect the parameter to be of the proper type.
```
So the binding result will be:
```
// BindEvent represents "Bind" event emitted by the contract.
type BindEvent struct {
	User []byte
	Keys keys.PublicKeys
}
```
The compiler will emit the conversion code to convert `Keys` to map and execution of the contract will fail.

>  the result will contain real type of the event parameters

This happens because we firstly check the predicted emitted parameter types against the configured one. The predicted emitted parameter type is `Map` in this case, because conversion code is emitted for this parameter. Thus, the check agains config is passed. After that if all parameter checks are passed we emit real parameter type to the binding configuration, and the real type in this case is `keys.PublicKeys`. As a result, we've got it in the resulting binding. I'm not sure whether we need it to be fixed right now, because if we're to solve https://github.com/nspcc-dev/neo-go/issues/3027, then the parameter check won't pass for such event and compilation will fail (because the baes type for `keys.PublicKeys` is an array, not `Map`).

AnnaShaleva: The perfect solution would be to store the smartcontract's real type of the parameter (not the one that we'll get after conversion), and if it differs from the config, then take the one that is declared in the config and generate RPC binding with it (do not pay attention to `RealType` in this case). It will work even without https://github.com/nspcc-dev/neo-go/issues/3027 fix. Implement?

AnnaShaleva: Fixed.

AnnaShaleva:
@roman-khimov, do we need binding files to be stored in this repo or it's OK to add them to .gitignore?

README.md
```diff
@@ -37,28 +37,54 @@ Sidechain contracts:
...
 
 To build and compile smart contract, run `make all` command. Compiled contracts
 `*_contract.nef` and manifest `config.json` files are placed in the 
-corresponding directories. 
+corresponding directories. Generated RPC binding files `rpcbinding.json` are
+placed in the corresponding `rpcbinding` directories. 
 
 ```
 $ make all
-/home/user/go/bin/cli contract compile -i alphabet -c alphabet/config.yml -m alphabet/config.json -o alphabet/alphabet_contract.nef
```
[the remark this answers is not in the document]
AnnaShaleva: Fixed.

Makefile
```diff
@@ -31,7 +31,9 @@ nns_sc = nns
 
 define sc_template
 $(2)$(1)/$(1)_contract.nef: $(2)$(1)/$(1)_contract.go
-	$(NEOGO) contract compile -i $(2)$(1) -c $(if $(2),$(2),$(1)/)config.yml -m $(2)$(1)/config.json -o $(2)$(1)/$(1)_contract.nef
+	$(NEOGO) contract compile -i $(2)$(1) -c $(if $(2),$(2),$(1)/)config.yml -m $(2)$(1)/config.json -o $(2)$(1)/$(1)_contract.nef --bindings $(2)$(1)/bindings_config.yml
+	mkdir -p rpc/$(1)
+	$(NEOGO) contract generate-rpcwrapper -o rpc/$(1)/rpcbinding.go -m $(2)$(1)/config.json --config $(2)$(1)/bindings_config.yml
```
roman-khimov, line 36: This, btw, should be handled by the Makefile. Makefile is not a shell script, it's all about targets, dependencies and the way they're produced.

.gitignore
```diff
@@ -1,6 +1,8 @@
 *.avm
 *.nef
 *.manifest.json
+bindings_config.yml
+rpcbinding.go
```
roman-khimov, line 4: `rpcbinding.go` files will be added eventually.

README.md
```diff
@@ -37,28 +37,21 @@ Sidechain contracts:
...
 
--   [neo-go](https://github.com/nspcc-dev/neo-go) >= 0.99.2
+-   [neo-go](https://github.com/nspcc-dev/neo-go) >= 0.102.0
 
 ## Compilation
 
 To build and compile smart contract, run `make all` command. Compiled contracts
 `*_contract.nef` and manifest `config.json` files are placed in the 
-corresponding directories. 
+corresponding directories. Generated RPC binding files `rpcbinding.json` are
```
roman-khimov, line 46: `rpcbinding.json`?

Makefile
AnnaShaleva: That's the only question I wasn't able to fix. Could you, please, review one more time and point out how to create a proper make target depending on all these `rpc/...` bindings?

README.md
AnnaShaleva: Typo, fixed.

Makefile
roman-khimov: Let's leave it for now.

roman-khimov:
@smallhive, can you try importing packages from this branch?

smallhive:
> @smallhive, can you try importing packages from this branch?

Working on it

rpc/audit/rpcbinding.go
```diff
@@ -0,0 +1,1591 @@
...
+
+// CommonBallot is a contract-specific common.Ballot type used by its methods.
+type CommonBallot struct {
+	ID []byte
+	Voters keys.PublicKeys
+	Height *big.Int
+}
+
+// LedgerBlock is a contract-specific ledger.Block type used by its methods.
+type LedgerBlock struct {
```
cthulhu-rider, line 33: we have this and similar called types in each package of the same level. Could u pls explain why do we need them to be exported? 

btw there are also more contract-specific ones like `AuditauditHeader`.

roman-khimov: It's nspcc-dev/neo-go#3071, they'll be removed once it's solved. We can live with some garbage in these for now.

Makefile
```diff
@@ -7,7 +7,7 @@ export GOBIN ?= $(shell pwd)/bin
 export CGO_ENABLED=0
 NEOGO ?= $(GOBIN)/cli
 VERSION ?= $(shell git describe --tags --dirty --match "v*" --always --abbrev=8 2>/dev/null || cat VERSION 2>/dev/null || echo "develop")
-NEOGOORIGMOD = github.com/nspcc-dev/neo-go
+NEOGOORIGMOD = github.com/nspcc-dev/neo-go@v0.101.5-0.20230802075307-afa4530c7d73
```
carpawell, line 10: @AnnaShaleva, since that is a `temporary use`, do we need an issue for that?

roman-khimov: There will be an issue to upgrade to 0.102.0 anyway. Can be created now, doesn't matter much.

PR #347, opened 2023-08-03
Add NNS record types to the binding

roman-khimov:


smallhive (approved):

.gitignore
```diff
@@ -1,5 +1,6 @@
 *.avm
 *.nef
+*~
```
carpawell, line 3: wow. what is it used for?

nns/recordtype.go
```diff
@@ -3,6 +3,9 @@ package nns
 // RecordType is domain name service record types.
 type RecordType byte
 
+// WARNING!
```
carpawell, line 6: can we move that into `const`? practice says devs look max 2 lines above when fixing/extending smth

carpawell (approved):

.gitignore
roman-khimov: File backup. I thought every .gitignore has this pattern.

nns/recordtype.go
roman-khimov: I've specifically tried to avoid having it in godoc, since it's only relevant for developers. The file is rather small and not changed often.

AnnaShaleva (commented):
LGTM.

PR #356, opened 2023-09-18
`go fmt` generated code

cthulhu-rider:
btw we can go further and use `goimports` tool since imports are unsorted, but i decided to not do that for now

with these changes `make fmt` (`go fmt ./...`) becomes no-op

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changelog for NeoFS Contract
 - RPC bindings generation (#345)
 
 ### Updated
+- `neo-go` to `v0.103.1`
 
 ### Removed
 - old unused (notary-disabled) events (#341)
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Sidechain contracts:
 
 To compile smart contracts you need:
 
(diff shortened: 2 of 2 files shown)


roman-khimov (commented):
Do we really need it? These are used for imports and doc generation, they're not supposed to be looked at.

cthulhu-rider:
> Do we really need it? These are used for imports and doc generation, they're not supposed to be looked at.

not a big problem, but clean `go fmt ./...` out is nice imo

maybe this will inspire generator to use https://pkg.go.dev/go/format@go1.18#Source

roman-khimov:
Huh, that's exactly what I was looking for. But for some reason `gofmt` works a bit differently, or maybe I wasn't looking deep enough. Yes, I think we better do this at the generator level (just in case anyone still wants to have a look), please file an issue.

carpawell:
Not needed after https://github.com/nspcc-dev/neo-go/pull/3164?

roman-khimov:
Likely so.

cthulhu-rider:
updated prerequisities with corresponding Neo Go version

roman-khimov (approved):

PR #357, opened 2023-09-21
NNS migration to 0.18.0

roman-khimov:


common/version.go
```diff
@@ -5,7 +5,7 @@ import "github.com/nspcc-dev/neo-go/pkg/interop/native/std"
 const (
 	major = 0
 	minor = 17
-	patch = 0
+	patch = 5
```
AnnaShaleva, line 8: Don't we need to update correspondingly the `prevMinor` and `prevPatch` as far? I'm not so deep in this part of contract updates, so maybe we don't need to.

nns/migration_test.go
```diff
@@ -0,0 +1,86 @@
...
+		require.NotEqual(t, 0, bal.Int64())
+		balances = append(balances, bal.Int64())
+	}
+
+	c.CheckUpdateSuccess(t)
+
+	for i, tld := range tlds {
+		// There is no owner after the upgrade.
+		_, err := c.TestInvoke(t, "ownerOf", tld)
+		require.Error(t, err)
```
AnnaShaleva, line 72: Could we check the exact error/error text? Just in case. ("token not found")

AnnaShaleva (commented):
LGTM otherwise.

CHANGELOG.md
```diff
@@ -4,16 +4,22 @@ Changelog for NeoFS Contract
 ## [Unreleased]
 
 ### Added
+- EACL validation in container.setEACL (#330)
+- Contract storage model documentation (#320)
```
carpawell, line 8: how do you define the capitalization of the first word? (my vote is for a capital one, but at least they should be synced)

common/version.go
carpawell: btw, why not `1`?

nns/migration_test.go
```diff
@@ -0,0 +1,86 @@
+package nns_test
+
+import (
+	"bytes"
+	"testing"
+
+	//	"github.com/mr-tron/base58"
```
carpawell, line 7: what does it mean?

nns/nns_contract.go
```diff
@@ -93,7 +100,33 @@ func Update(nef []byte, manifest string, data interface{}) {
...
+			iter = storage.Find(ctx, []byte{prefixName}, storage.DeserializeValues)
+			rec  nsIteratorValue
+			ser  []byte
+		)
+		for iterator.Next(iter) {
+			rec = iterator.Value(iter).(nsIteratorValue)
+			if std.MemorySearch([]byte(rec.v.Name), []byte{'.'}) != -1 { // Non-TLD
+				continue
+			}
+			runtime.Log("up " + std.Base64Encode(rec.v.Owner))
```
carpawell, line 123: "up"?

common/version.go
roman-khimov: `Prev` constants are OK as is, it's the oldest version we support upgrading from and so far we have not removed any of the migration code. The value is somewhere in the middle of 17.0 and 18.0. 1 makes sense in general, but it doesn't matter much here since there are new APIs.

nns/migration_test.go
roman-khimov: Sure, fixed.

CHANGELOG.md
roman-khimov: Looks nice here.

nns/migration_test.go
roman-khimov: Copy&paste detected. Fixed.

nns/nns_contract.go
roman-khimov: Debug, dropped.

AnnaShaleva (approved):

CHANGELOG.md
carpawell: what does "Contract" mean here? why is it "Contract" but the others are "ability", "bump" and so on?

cthulhu-rider (approved):

CHANGELOG.md
roman-khimov: Added... some kind of documentation.

carpawell (approved):

PR #360, opened 2023-09-21
container: add alias method, fix #343

### LINKED ISSUE #343: Method to get container nice name by ID
It's easy to resolve ID by name with NNS, but it's not so easy to get a name by ID, even though internally container contract has this data. So it should provide a (safe) method for this data to be accessible from outside.

roman-khimov:


rpc/container/rpcbinding.go
```diff
@@ -241,6 +241,11 @@ func New(actor Actor, hash util.Uint160) *Contract {
 	return &Contract{ContractReader{actor, hash}, actor, hash}
 }
 
+// Alias invokes `alias` method of contract.
+func (c *ContractReader) Alias(cid []byte) ([]byte, error) {
+	return unwrap.Bytes(c.invoker.Call(c.hash, "alias", cid))
```
AnnaShaleva, line 246: `unwrap.Bytes` won't be able to handle `Null` stackitem and will return an error. Is it the expected binding behaviour?

tests/container_test.go
```diff
@@ -196,6 +196,7 @@ func TestContainerPut(t *testing.T) {
 		})
 		cNNS := c.CommitteeInvoker(nnsHash)
 		cNNS.Invoke(t, expected, "resolve", "mycnt.neofs", int64(nns.TXT))
+		c.Invoke(t, stackitem.NewBuffer([]byte("mycnt.neofs")), "alias", cnt.id[:])
```
AnnaShaleva, line 199: Could we add a test for missing alias to check the Null result?

cthulhu-rider (dismissed):

rpc/container/rpcbinding.go
roman-khimov: It's expected. We can have `string` here, given the current state of `unwrap` that might be a bit more appropriate. But we couldn't have `nil, nil` return in the future this way. But the usefulness of `nil, nil` is questionable as well.

tests/container_test.go
roman-khimov: Sure, there you have it.

rpc/container/rpcbinding.go
roman-khimov: Let's try strings.

AnnaShaleva (approved):

cthulhu-rider (approved):

PR #361, opened 2023-09-22
Simplify resolving NNS hashes via RPC binding

roman-khimov:


rpc/nns/hashes.go
```diff
@@ -0,0 +1,62 @@
...
+	return c.Hash, nil
+}
+
+// ResolveFSContract is a convenience method that doesn't exist in the NNS
+// contract itself (it doesn't care which data is stored there). It assumes
+// that contracts follow the [ContractTLD] convention, gets simple contract
+// names (like "container" or "netmap") and extracts the hash for the
+// respective NNS record in any of the formats (of which historically there's
+// been a few).
+func (c *ContractReader) ResolveFSContract(name string) (util.Uint160, error) {
```
smallhive, line 45: It would be perfect to have some enum for "container", "netmap" and other possible contract names.
If it already exists, the reference in the comment will be useful

rpc/nns/hashes.go
```diff
@@ -27,3 +35,28 @@ func InferHash(sg ContractStateGetter) (util.Uint160, error) {
...
 }
+
+// ResolveFSContract is a convenience method that doesn't exist in the NNS
+// contract itself (it doesn't care which data is stored there). It assumes
+// that contracts follow the [ContractTLD] convention, gets simple contract
+// names (like "container" or "netmap") and extracts the hash for the
+// respective NNS record in any of the formats (of which historically there's
+// been a few).
+func (c *ContractReader) ResolveFSContract(name string) (util.Uint160, error) {
+	strs, err := c.Resolve(name+"."+ContractTLD, TXT)
```
AnnaShaleva, line 46: IIRC, NNS's "resolve" method is able to return `stackitem.Null`, and `c.Resolve` isn't able to properly handle this. Don't we need to add some comment about it? Although the `ResolveFSContract` return value is either hash or error, and this behaviour should be OK. But maybe we also need to check some other usages of `c.Resolce`.

rpc/nns/hashes.go
```diff
@@ -27,3 +35,28 @@ func InferHash(sg ContractStateGetter) (util.Uint160, error) {
...
+		h, err := util.Uint160DecodeStringLE(strs[i])
+		if err == nil {
+			return h, nil
+		}
+
+		h, err = address.StringToUint160(strs[i])
+		if err == nil {
+			return h, nil
+		}
+	}
```
AnnaShaleva, line 60: Is it ever possible that the first record doesn't contain the desired hash in the LE format? I thought that it's always true for all contract: first record is LE string hash, second record is the same hash in the address format.

rpc/nns/hashes.go
roman-khimov: > eturn value is either hash or error

That's exactly why it's the the way it is now.

rpc/nns/hashes.go
roman-khimov: It can be just one (any) of them and I'd like to move to NEP-proposed scheme in the future.

rpc/nns/hashes.go
roman-khimov: It doesn't, at least not in some common space (various utilities like adm do know about them).

smallhive (approved):

rpc/nns/hashes.go
```diff
@@ -0,0 +1,29 @@
+package nns
+
+import (
+	"github.com/nspcc-dev/neo-go/pkg/core/state"
+	"github.com/nspcc-dev/neo-go/pkg/util"
+)
+
+// ID is the default NNS contract ID in all NeoFS networks. NeoFS networks
+// always deploy NNS first and can't work without it, therefore it always gets
+// and ID of 1.
```
carpawell, line 13: some typo here?

rpc/nns/hashes.go
carpawell: cant we add some constants? as a user i would be more glad not to figure names myself

roman-khimov: OK, let's try them this way.

rpc/nns/hashes.go
roman-khimov: Fixed.

rpc/nns/names.go
```diff
@@ -0,0 +1,13 @@
+package nns
+
+// A set of standard contract names deployed into NeoFS sidechain.
+const (
```
AnnaShaleva, line 4: We have a set of names in neofs-node repo (https://github.com/nspcc-dev/neofs-node/blob/920743715ddfcfd882011f5924c5476bf9e940bd/cmd/neofs-adm/internal/modules/morph/initialize_deploy.go#L39-L49). Maybe it's a good idea to remove them from neofs-node and use these constants instead?

roman-khimov: Where do you think I've copy-pasted them from?! Of course it should be done during upgrade to 0.18.0.

rpc/nns/names.go
```diff
@@ -0,0 +1,13 @@
...
+	NameAlphabet   = "alphabet"
```
cthulhu-rider, line 5: Alphabet contracts are currently special cuz they are named with indices `alphabet0`, `alphabet1`, etc.

cthulhu-rider (commented):
good candidate to provide code example like
```go
// Resolve addresses of the NeoFS smart contracts deployed in the particular
// NeoFS Sidechain by their NNS domain names.
func ExampleContractReader_ResolveFSContract() {
	const sidechainRPCEndpoint = "https://rpc1.morph.fs.neo.org:40341"

	c, err := rpcclient.New(context.Background(), sidechainRPCEndpoint, rpcclient.Options{})
	if err != nil {
		log.Fatal(err)
	}

	err = c.Init()
	if err != nil {
		log.Fatal(err)
	}

	nnsAddress, err := nns.InferHash(c)
	if err != nil {
		log.Fatal(err)
	}

	nnsContract := nns.NewReader(invoker.New(c, nil), nnsAddress)

	for _, name := range []string{
		nns.NameAudit,
		nns.NameBalance,
		nns.NameContainer,
		nns.NameNeoFSID,
		nns.NameNetmap,
		nns.NameProxy,
		nns.NameReputation,
	} {
		addr, err := nnsContract.ResolveFSContract(name)
		if err != nil {
			log.Fatal(err)
		}

		fmt.Printf("%s: %s\n", name, addr)
	}
}
```

rpc/nns/names.go
roman-khimov: Maybe it should be `NameAlphabetPrefix` then?

cthulhu-rider: yeah, such name gonna be much more correct

roman-khimov: Fixed.

cthulhu-rider (approved):

AnnaShaleva (approved):

carpawell (approved):

PR #366, opened 2023-11-09
nns: fix expired TLD registration

roman-khimov:
If a TLD has expired:
 * it can't be renewed because of "name has expired"
 * it can't be registered again because of "TLD already exists"

The first one is OK, but the second one is fixed here.

### DIFF
--- a/tests/nns_test.go
+++ b/tests/nns_test.go
@@ -246,6 +246,8 @@ func TestExpiration(t *testing.T) {
 	c := newNNSInvoker(t, true)
 
 	refresh, retry, expire, ttl := int64(101), int64(102), int64(msPerYear/1000*10), int64(104)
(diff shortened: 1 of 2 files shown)


tests/nns_test.go
```diff
@@ -278,6 +280,13 @@ func TestExpiration(t *testing.T) {
 
 	c.InvokeFail(t, "name has expired", "getAllRecords", "testdomain.com")
 	c.InvokeFail(t, "name has expired", "ownerOf", "testdomain.com")
+
+	c.InvokeFail(t, "name has expired", "renew", "newtld")
```
carpawell, line 284: wow so many tests inside a singe `Test*` without any `t.Run`

carpawell (approved):

PR #368, opened 2023-11-20
feat/Make `NewEpoch` callback mechanism more generic

carpawell:
Do not hardcode container and balance contract as a `NewEpoch` events receivers in the netmap contract. Use the netmap's registration method on a deploy stage instead.

### DIFF
--- a/balance/balance_contract.go
+++ b/balance/balance_contract.go
@@ -78,6 +78,8 @@ func _deploy(data any, isUpdate bool) {
 		return
 	}
 
+	common.SubscribeForNewEpoch()
+
 	runtime.Log("balance contract initialized")
 }
 
--- a/common/netmap.go
+++ b/common/netmap.go
@@ -1,0 +1,39 @@
+package common
+
+import (
+	"github.com/nspcc-dev/neo-go/pkg/interop"
+	"github.com/nspcc-dev/neo-go/pkg/interop/contract"
+	"github.com/nspcc-dev/neo-go/pkg/interop/convert"
+	"github.com/nspcc-dev/neo-go/pkg/interop/lib/address"
+	"github.com/nspcc-dev/neo-go/pkg/interop/native/management"
+	"github.com/nspcc-dev/neo-go/pkg/interop/native/std"
+	"github.com/nspcc-dev/neo-go/pkg/interop/runtime"
+)
+
+// SubscribeForNewEpoch registers calling contract as a NewEpoch
+// callback requester. Netmap contract's address is taken from the
+// NNS contract, therefore, it must be presented and filled with
+// netmap information for a correct SubscribeForNewEpoch call; otherwise
+// a successive call is not guaranteed.
+// Caller must have `NewEpoch` method with a single numeric argument.
+func SubscribeForNewEpoch() {
+	nnsContract := management.GetContractByID(1)
+	if nnsContract == nil {
+		panic("missing NNS contract")
+	}
+
+	resolveRes := contract.Call(nnsContract.Hash, "resolve", contract.ReadOnly, "netmap.neofs", 16)
+	records := resolveRes.([]string)
+	if len(records) == 0 {
+		panic("did not find a record of the Netmap contract in the NNS")
+	}
+
+	var netmapContract interop.Hash160
+	if len(records[0]) == 2*interop.Hash160Len {
+		netmapContract = convert.ToBytes(std.Atoi(records[0], 16))
+	} else {
+		netmapContract = address.ToHash160(records[0])
+	}
+
+	contract.Call(netmapContract, "subscribeForNewEpoch", contract.All, runtime.GetExecutingScriptHash())
+}
--- a/common/netmap.go
+++ b/common/netmap.go
@@ -1,12 +1,7 @@
 package common
 
 import (
-	"github.com/nspcc-dev/neo-go/pkg/interop"
 	"github.com/nspcc-dev/neo-go/pkg/interop/contract"
-	"github.com/nspcc-dev/neo-go/pkg/interop/convert"
-	"github.com/nspcc-dev/neo-go/pkg/interop/lib/address"
-	"github.com/nspcc-dev/neo-go/pkg/interop/native/management"
-	"github.com/nspcc-dev/neo-go/pkg/interop/native/std"
 	"github.com/nspcc-dev/neo-go/pkg/interop/runtime"
 )
 
@@ -17,23 +12,6 @@ import (
 // a successive call is not guaranteed.
 // Caller must have `NewEpoch` method with a single numeric argument.
 func SubscribeForNewEpoch() {
-	nnsContract := management.GetContractByID(1)
-	if nnsContract == nil {
-		panic("missing NNS contract")
-	}
-
-	resolveRes := contract.Call(nnsContract.Hash, "resolve", contract.ReadOnly, "netmap.neofs", 16)
-	records := resolveRes.([]string)
-	if len(records) == 0 {
-		panic("did not find a record of the Netmap contract in the NNS")
-	}
-
-	var netmapContract interop.Hash160
-	if len(records[0]) == 2*interop.Hash160Len {
-		netmapContract = convert.ToBytes(std.Atoi(records[0], 16))
-	} else {
-		netmapContract = address.ToHash160(records[0])
-	}
-
+	netmapContract := ResolveContractHash("netmap")
 	contract.Call(netmapContract, "subscribeForNewEpoch", contract.All, runtime.GetExecutingScriptHash())
 }
--- a/common/nns.go
+++ b/common/nns.go
@@ -1,0 +1,43 @@
+package common
+
+import (
+	"github.com/nspcc-dev/neo-go/pkg/interop"
+	"github.com/nspcc-dev/neo-go/pkg/interop/contract"
+	"github.com/nspcc-dev/neo-go/pkg/interop/convert"
+	"github.com/nspcc-dev/neo-go/pkg/interop/lib/address"
+	"github.com/nspcc-dev/neo-go/pkg/interop/native/management"
+	"github.com/nspcc-dev/neo-go/pkg/interop/native/std"
+)
+
+// ResolveContractHash resolves contract hash by its well-known NeoFS name.
+// Contract name should be lowercased, should not include `.neofs` TLD. Example
+// values: "netmap", "container", etc.
+// Relies on some NeoFS specifics:
+//  1. NNS contract should be deployed first (and should have `1` contract ID)
+//  2. It should be prefilled with contract hashes by their names (no
+//     capitalized chars; no `neofs` TLD)
+func ResolveContractHash(contractName string) interop.Hash160 {
+	// get NNS contract (it always has ID=1 in the NeoFS Sidechain)
+	nnsContract := management.GetContractByID(1)
+	if nnsContract == nil {
+		panic("missing NNS contract")
+	}
+
+	resResolve := contract.Call(nnsContract.Hash, "resolve", contract.ReadOnly,
+		contractName+".neofs", 16, // TXT
+	)
+
+	records := resResolve.([]string)
+	if len(records) == 0 {
+		panic("did not find a record of the " + contractName + " contract in the NNS")
+	}
+
+	var hash interop.Hash160
+	if len(records[0]) == 2*interop.Hash160Len {
+		hash = convert.ToBytes(std.Atoi(records[0], 16))
+	} else {
+		hash = address.ToHash160(records[0])
+	}
+
+	return hash
+}
--- a/container/config.yml
+++ b/container/config.yml
@@ -2,7 +2,7 @@ name: "NeoFS Container"
 safemethods: ["alias", "count", "containersOf", "get", "owner", "list", "eACL", "getContainerSize", "listContainerSizes", "iterateContainerSizes", "iterateAllContainerSizes", "version"]
 permissions:
   - methods: ["update", "addKey", "transferX",
-              "register", "registerTLD", "addRecord", "deleteRecords"]
+              "register", "registerTLD", "addRecord", "deleteRecords", "subscribeForNewEpoch"]
 events:
   - name: PutSuccess
     parameters:
--- a/container/container_contract.go
+++ b/container/container_contract.go
@@ -203,7 +203,7 @@ func switchToNotary(ctx storage.Context) {
 // nolint:deadcode,unused
 func registerNiceNameTLD(addrNNS interop.Hash160, nnsRoot string) {
 	isAvail := contract.Call(addrNNS, "isAvailable", contract.AllowCall|contract.ReadStates,
-		"container").(bool)
+		nnsRoot).(bool)
 	if !isAvail {
 		return
 	}
--- a/container/container_contract.go
+++ b/container/container_contract.go
@@ -169,6 +169,8 @@ func _deploy(data any, isUpdate bool) {
 	// add NNS root for container alias domains
 	registerNiceNameTLD(args.addrNNS, args.nnsRoot)
 
+	common.SubscribeForNewEpoch()
+
 	runtime.Log("container contract initialized")
 }
 
--- a/netmap/config.yml
+++ b/netmap/config.yml
@@ -17,3 +17,7 @@ events:
     parameters:
       - name: epoch
         type: Integer
+  - name: NewEpochSubscription
+    parameters:
+      - name: contract
+        type: Hash160
--- a/netmap/migration_test.go
+++ b/netmap/migration_test.go
@@ -1,6 +1,7 @@
 package netmap_test
 
 import (
+	"bytes"
 	"testing"
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
@@ -24,19 +25,48 @@ func TestMigration(t *testing.T) {
 	require.NoError(t, err)
 }
 
-var notaryDisabledKey = []byte("notary")
+var (
+	notaryDisabledKey = []byte("notary")
+
+	newEpochSubsNewPrefix = []byte("e")
+	containerHashOldKey   = []byte("containerScriptHash")
+	balanceHashOldKey     = []byte("balanceScriptHash")
+)
 
 func testMigrationFromDump(t *testing.T, d *dump.Reader) {
+	var containerHash util.Uint160
+	var balanceHash util.Uint160
+	var err error
+
 	// init test contract shell
-	c := migration.NewContract(t, d, "netmap", migration.ContractOptions{})
+	c := migration.NewContract(t, d, "netmap", migration.ContractOptions{
+		StorageDumpHandler: func(key, value []byte) {
+			if bytes.Equal(containerHashOldKey, key) {
+				containerHash, err = util.Uint160DecodeBytesLE(value)
+				require.NoError(t, err)
+
+				return
+			}
+
+			if bytes.Equal(balanceHashOldKey, key) {
+				balanceHash, err = util.Uint160DecodeBytesLE(value)
+				require.NoError(t, err)
+
+				return
+			}
+		},
+	})
+
+	require.NotZerof(t, balanceHash, "missing storage item %q with Balance contract address", balanceHashOldKey)
+	require.NotZerof(t, containerHash, "missing storage item %q with Container contract address", containerHashOldKey)
 
 	var notary bool
 	updPrm := []any{
 		!notary,
-		util.Uint160{},  // Balance contract
-		util.Uint160{},  // Container contract
-		[]any{}, // Key list, unused
-		[]any{}, // Config
+		util.Uint160{}, // Balance contract
+		util.Uint160{}, // Container contract
+		[]any{},        // Key list, unused
+		[]any{},        // Config
 	}
 
 	migration.SkipUnsupportedVersions(t, c, updPrm...)
@@ -157,6 +187,8 @@ func testMigrationFromDump(t *testing.T, d *dump.Reader) {
 
 	c.CheckUpdateSuccess(t, updPrm...)
 
+	checkNewEpochSubscribers(t, c, balanceHash, containerHash)
+
 	// check that contract was updates as expected
 	newPendingVotes := readPendingVotes()
 	newVersion := readVersion()
@@ -183,4 +215,37 @@ func testMigrationFromDump(t *testing.T, d *dump.Reader) {
 		require.True(t, ok)
 		require.ElementsMatch(t, vPrev, vNew, "%d-th past netmap snapshot should remain", k)
 	}
+}
+
+func checkNewEpochSubscribers(t *testing.T, contract *migration.Contract, balanceWant, containerWant util.Uint160) {
+	require.Nil(t, contract.GetStorageItem(balanceHashOldKey))
+	require.Nil(t, contract.GetStorageItem(containerHashOldKey))
+
+	// contracts are migrated in alphabetical order at least for now
+
+	var balanceMigrated bool
+	var containerMigrated bool
+
+	contract.SeekStorage(append(newEpochSubsNewPrefix, 0), func(k, v []byte) bool {
+		balanceGot, err := util.Uint160DecodeBytesLE(k)
+		require.NoError(t, err)
+		require.Equal(t, balanceWant, balanceGot)
+
+		balanceMigrated = true
+
+		return true
+	})
+
+	contract.SeekStorage(append(newEpochSubsNewPrefix, 1), func(k, v []byte) bool {
+		containerGot, err := util.Uint160DecodeBytesLE(k)
+		require.NoError(t, err)
+		require.Equal(t, containerWant, containerGot)
+
+		containerMigrated = true
+
+		return true
+	})
+
+	require.True(t, balanceMigrated, "balance contact hash migration")
+	require.True(t, containerMigrated, "container contact hash migration")
 }
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -73,7 +73,8 @@ const (
 	containerContractKey = "containerScriptHash"
 	balanceContractKey   = "balanceScriptHash"
 
-	cleanupEpochMethod = "newEpoch"
+	newEpochSubscribersPrefix = "e"
+	cleanupEpochMethod        = "newEpoch"
 
 	// nodeKeyOffset is an offset in a serialized node info representation (V2 format)
 	// marking the start of the node's public key.
@@ -95,8 +96,8 @@ func _deploy(data any, isUpdate bool) {
 
 	var args = data.(struct {
 		notaryDisabled bool
-		addrBalance    interop.Hash160
-		addrContainer  interop.Hash160
+		_              interop.Hash160 // Balance contract not used legacy
+		_              interop.Hash160 // Container contract not used legacy
 		keys           []interop.PublicKey
 		config         [][]byte
 		version        int
@@ -162,11 +163,19 @@ func _deploy(data any, isUpdate bool) {
 			switchToNotary(ctx)
 		}
 
-		return
-	}
+		if args.version < 19_000 {
+			balanceContract := storage.Get(ctx, balanceContractKey).(interop.Hash160)
+			key := append([]byte(newEpochSubscribersPrefix), append([]byte{byte(0)}, balanceContract...)...)
+			storage.Put(ctx, key, []byte{})
+			storage.Delete(ctx, balanceContractKey)
 
-	if len(args.addrBalance) != interop.Hash160Len || len(args.addrContainer) != interop.Hash160Len {
-		panic("incorrect length of contract script hash")
+			containerContract := storage.Get(ctx, containerContractKey).(interop.Hash160)
+			key = append([]byte(newEpochSubscribersPrefix), append([]byte{byte(1)}, containerContract...)...)
+			storage.Put(ctx, key, []byte{})
+			storage.Delete(ctx, containerContractKey)
+		}
+
+		return
 	}
 
 	// epoch number is a little endian int, it doesn't need to be serialized
@@ -179,9 +188,6 @@ func _deploy(data any, isUpdate bool) {
 		common.SetSerialized(ctx, append(prefix, byte(i)), []Node{})
 	}
 	storage.Put(ctx, snapshotCurrentIDKey, 0)
-
-	storage.Put(ctx, balanceContractKey, args.addrBalance)
-	storage.Put(ctx, containerContractKey, args.addrContainer)
 
 	runtime.Log("netmap contract initialized")
 }
@@ -601,6 +607,32 @@ func ListConfig() []record {
 	return config
 }
 
+// SubscribeForNewEpoch registers passed contract as a NewEpoch event
+// subscriber. Such a contract must have a `NewEpoch` method with a single
+// numeric parameter. Transactions that call SubscribeForNewEpoch must be
+// witnessed by the Alphabet.
+// Produces `NewEpochSubscription` notification event with a just
+// registered recipient in a success case.
+func SubscribeForNewEpoch(contract interop.Hash160) {
+	common.CheckAlphabetWitness(common.AlphabetAddress())
+
+	ctx := storage.GetContext()
+	var num byte
+	it := storage.Find(ctx, []byte(newEpochSubscribersPrefix), storage.None)
+	for iterator.Next(it) {
+		num += 1
+	}
+
+	var key []byte
+	key = append(key, []byte(newEpochSubscribersPrefix)...)
+	key = append(key, num)
+	key = append(key, contract...)
+
+	storage.Put(ctx, key, []byte{})
+
+	runtime.Notify("NewEpochSubscription", contract)
+}
+
 // Version returns the version of the contract.
 func Version() int {
 	return common.Version
@@ -685,9 +717,9 @@ func setConfig(ctx storage.Context, key, val any) {
 }
 
 func cleanup(ctx storage.Context, epoch int) {
-	balanceContractAddr := storage.Get(ctx, balanceContractKey).(interop.Hash160)
-	contract.Call(balanceContractAddr, cleanupEpochMethod, contract.All, epoch)
-
-	containerContractAddr := storage.Get(ctx, containerContractKey).(interop.Hash160)
-	contract.Call(containerContractAddr, cleanupEpochMethod, contract.All, epoch)
+	it := storage.Find(ctx, newEpochSubscribersPrefix, storage.RemovePrefix|storage.KeysOnly)
+	for iterator.Next(it) {
+		contractHash := interop.Hash160(iterator.Value(it).([]byte)[1:]) // one byte is for number prefix
+		contract.Call(contractHash, cleanupEpochMethod, contract.All, epoch)
+	}
 }
--- a/netmap/netmap_contract.go
+++ b/netmap/netmap_contract.go
@@ -71,8 +71,10 @@ const (
 	snapshotEpoch        = "snapshotEpoch"
 	snapshotBlockKey     = "snapshotBlock"
 
+	// nolint:unused // used only in _deploy func which is nolinted too
 	containerContractKey = "containerScriptHash"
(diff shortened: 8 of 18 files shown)


netmap/netmap_contract.go
```diff
@@ -601,6 +602,23 @@ func ListConfig() []record {
 	return config
 }
 
+// RegisterNewEpochRecipient registers passed contract as a NewEpoch event
+// receiver. Such a contract must have a `NewEpoch` method with a single numeric
+// argument. Transactions that call RegisterNewEpochRecipient must be witnessed
+// by the Alphabet.
+// Produces `NewEpochHookRegistration` notification event with a just registered
+// recipient in a success case.
+func RegisterNewEpochRecipient(contract interop.Hash160) {
```
roman-khimov, line 611: Do you need to pass `contract`? Likely you can take caller script hash.

netmap/netmap_contract.go
```diff
@@ -685,9 +703,9 @@ func setConfig(ctx storage.Context, key, val any) {
 }
 
 func cleanup(ctx storage.Context, epoch int) {
-	balanceContractAddr := storage.Get(ctx, balanceContractKey).(interop.Hash160)
-	contract.Call(balanceContractAddr, cleanupEpochMethod, contract.All, epoch)
-
-	containerContractAddr := storage.Get(ctx, containerContractKey).(interop.Hash160)
-	contract.Call(containerContractAddr, cleanupEpochMethod, contract.All, epoch)
+	it := storage.Find(ctx, newEpochRecipientsKey, storage.RemovePrefix|storage.KeysOnly)
```
roman-khimov, line 706: Does order of invocations matter here?

netmap/config.yml
```diff
@@ -17,3 +17,7 @@ events:
     parameters:
       - name: epoch
         type: Integer
+  - name: NewEpochHookRegistration
+    parameters:
+      - name: registree
```
roman-khimov, line 22: Caller/contract?

balance/balance_contract.go
```diff
@@ -75,6 +75,8 @@ func _deploy(data any, isUpdate bool) {
 			switchToNotary(ctx)
 		}
 
+		common.RegisterNewEpochHook()
```
roman-khimov, line 78: So, how about deployments?

common/netmap.go
```diff
@@ -0,0 +1,39 @@
...
+func RegisterNewEpochHook() {
+	nnsContract := management.GetContractByID(1)
+	if nnsContract == nil {
+		panic("missing NNS contract")
+	}
+
+	resolveRes := contract.Call(nnsContract.Hash, "resolve", contract.ReadOnly, "netmap.neofs", 16)
+	records := resolveRes.([]string)
+	if len(records) == 0 {
+		panic("did not find a record of the Proxy contract in the NNS")
```
roman-khimov, line 28: Proxy?

common/netmap.go
```diff
@@ -0,0 +1,39 @@
...
+// netmap information for a correct RegisterNewEpochHook call; otherwise
+// a successive call is not guaranteed.
+// Caller must have `NewEpoch` method with a single numeric argument.
+func RegisterNewEpochHook() {
+	nnsContract := management.GetContractByID(1)
+	if nnsContract == nil {
+		panic("missing NNS contract")
+	}
+
+	resolveRes := contract.Call(nnsContract.Hash, "resolve", contract.ReadOnly, "netmap.neofs", 16)
```
roman-khimov, line 25: Contract bindings would be helpful here, but OK for now.

container/container_contract.go
```diff
@@ -138,6 +138,8 @@ func _deploy(data any, isUpdate bool) {
 			switchToNotary(ctx)
 		}
 
+		common.RegisterNewEpochHook()
```
roman-khimov, line 141: Deploy?

netmap/netmap_contract.go
```diff
@@ -685,9 +703,9 @@ func setConfig(ctx storage.Context, key, val any) {
 }
 
 func cleanup(ctx storage.Context, epoch int) {
-	balanceContractAddr := storage.Get(ctx, balanceContractKey).(interop.Hash160)
```
roman-khimov, line 688: We can migrate these in `netmap` and don't register on balance/container update. But I don't see any clear advantage either way.

netmap/netmap_contract.go
```diff
@@ -685,9 +703,9 @@ func setConfig(ctx storage.Context, key, val any) {
...
-	contract.Call(balanceContractAddr, cleanupEpochMethod, contract.All, epoch)
-
-	containerContractAddr := storage.Get(ctx, containerContractKey).(interop.Hash160)
```
roman-khimov, line 691: How about deleting these keys from DB on update and marking these options as unused and not requiring any checks on deploy?

roman-khimov (changes requested):
Tests fail.

netmap/netmap_contract.go
carpawell: missed that. but we need to decide: migration or removal

netmap/netmap_contract.go
carpawell: > We can migrate these in netmap and don't register on balance/container update.

too hard to control it IMO. how to decide when a contract needs to be registered if you are a stupid contract? or rely on migration and never ever register container and balance again?

common/netmap.go
carpawell: you mean neo-go has smth for contracts (not RPC) too? i know nothing about it

balance/balance_contract.go
carpawell: clarify your questions, please

netmap/config.yml
carpawell: `contract`

netmap/netmap_contract.go
carpawell: you mean Container -> Balance or Balance -> Container? no

netmap/netmap_contract.go
carpawell: > Likely you can take caller script hash

yes, that is what `common` pkg does in this PR. it is a kinda of feature that allows you to register some contracts to test or fix smth manually

balance/balance_contract.go
carpawell: clarify your questions, please

netmap/netmap_contract.go
carpawell: > Likely you can take caller script hash

yes, that is what `common` pkg does in this PR. it is a kinda of feature that allows you to register some contracts to test or fix smth manually

netmap/netmap_contract.go
carpawell: you mean Container -> Balance or Balance -> Container? no

netmap/config.yml
carpawell: `contract`

common/netmap.go
carpawell: fixed

common/netmap.go
carpawell: you mean neo-go has smth for contracts (not RPC) too? i know nothing about it

balance/balance_contract.go
carpawell: oh, wrong `if`, i got you

common/netmap.go
AnnaShaleva: Try `./bin/neo-go contract generate-wrapper`.

common/netmap.go
```diff
@@ -0,0 +1,39 @@
...
+	if len(records) == 0 {
+		panic("did not find a record of the Proxy contract in the NNS")
+	}
+
+	var netmapContract interop.Hash160
+	if len(records[0]) == 2*interop.Hash160Len {
+		netmapContract = convert.ToBytes(std.Atoi(records[0], 16))
+	} else {
+		netmapContract = address.ToHash160(records[0])
+	}
```
AnnaShaleva, line 36: This code is a part of alphabet contact as far, so maybe move it to a single method and reuse?

container/config.yml
```diff
@@ -2,7 +2,7 @@ name: "NeoFS Container"
 safemethods: ["alias", "count", "containersOf", "get", "owner", "list", "eACL", "getContainerSize", "listContainerSizes", "iterateContainerSizes", "iterateAllContainerSizes", "version"]
 permissions:
   - methods: ["update", "addKey", "transferX",
-              "register", "registerTLD", "addRecord", "deleteRecords"]
+              "register", "registerTLD", "addRecord", "deleteRecords", "registerNewEpochRecipient"]
```
AnnaShaleva, line 5: It may sound a little NeoGo-ish, but how about `subscribeForNewEpoch`? A lot of other `register` methods exist in NeoFS contracts.

netmap/netmap_contract.go
```diff
@@ -601,6 +602,23 @@ func ListConfig() []record {
...
+// by the Alphabet.
+// Produces `NewEpochHookRegistration` notification event with a just registered
+// recipient in a success case.
+func RegisterNewEpochRecipient(contract interop.Hash160) {
+	common.CheckAlphabetWitness(common.AlphabetAddress())
+
+	ctx := storage.GetContext()
+	key := append([]byte(newEpochRecipientsKey), contract...)
+
+	storage.Put(ctx, key, 1)
```
AnnaShaleva, line 617: Put an empty byte array, save a bit of storage!

netmap/netmap_contract.go
roman-khimov: Old keys should be removed, they're useless now. And given that they're a part of contract internal storage, they should be migrated into new ones.

netmap/netmap_contract.go
roman-khimov: Suppose you have updated netmap, but have not updated balance/container, what's gonna happen? In an ideal world, it all will work the same because netmap would migrate its callback thingies internally, not requiring old contracts to register somehow to continue their work.

common/netmap.go
roman-khimov: This will need to be done for all contracts, so for now we better leave it as is.

netmap/netmap_contract.go
roman-khimov: Can it become important in the future? We can register them with prefix_index_hash key, just in case.

carpawell: > Can it become important in the future?

anything can happen

> We can register them with prefix_index_hash

what exactly is `index` in your suggestion? registration number?

roman-khimov: Yeah, a simple byte will likely work, can't imagine 255+ contracts subscribing to this. At least there would be some predictability in the order, otherwise it's simply hash-dependent and will vary from network to network. Who knows how it can affect us in the future.

netmap/netmap_contract.go
carpawell: hm, did not know it was possible to store nothing, fixed

common/netmap.go
carpawell: done

container/config.yml
carpawell: goodish

netmap/netmap_contract.go
carpawell: done

netmap/netmap_contract.go
carpawell: doen

netmap/netmap_contract.go
carpawell: done

carpawell:
Ready one more time. Look carefully, please.

netmap/netmap_contract.go
```diff
@@ -73,7 +73,8 @@ const (
 	containerContractKey = "containerScriptHash"
```
carpawell, line 73: no idea, they are used

netmap/netmap_contract.go
```diff
@@ -95,8 +96,6 @@ func _deploy(data any, isUpdate bool) {
 
 	var args = data.(struct {
 		notaryDisabled bool
-		addrBalance    interop.Hash160
-		addrContainer  interop.Hash160
```
AnnaShaleva, line 99: So we don't care about compatibility with the old caller code here, right? The caller must know about this change and provide correct arguments, and if the caller's version is old, then deployment fails, and we're OK with it?

netmap/netmap_contract.go
```diff
@@ -162,11 +161,19 @@ func _deploy(data any, isUpdate bool) {
 			switchToNotary(ctx)
 		}
 
-		return
-	}
+		if args.version <= 19_000 {
+			balanceContract := storage.Get(ctx, balanceContractKey).(interop.Hash160)
+			key := append([]byte(newEpochSubscribersKey), append([]byte{byte(0)}, balanceContract...)...)
```
AnnaShaleva, line 166: You don't need `0` and `1` prefixes. Just use `newEpochSubscribersKey` plus contract hash, and it's exactly the unique key that you're looking for.

netmap/netmap_contract.go
```diff
@@ -601,6 +605,32 @@ func ListConfig() []record {
...
+// Produces `NewEpochHookRegistration` notification event with a just
+// registered recipient in a success case.
+func SubscribeForNewEpoch(contract interop.Hash160) {
+	common.CheckAlphabetWitness(common.AlphabetAddress())
+
+	ctx := storage.GetContext()
+	var num byte
+	it := storage.Find(ctx, []byte(newEpochSubscribersKey), storage.None)
+	for iterator.Next(it) {
+		num += 1
```
AnnaShaleva, line 621: And here, you don't need `num` as far.

netmap/netmap_contract.go
```diff
@@ -685,9 +715,9 @@ func setConfig(ctx storage.Context, key, val any) {
...
 
 func cleanup(ctx storage.Context, epoch int) {
-	balanceContractAddr := storage.Get(ctx, balanceContractKey).(interop.Hash160)
-	contract.Call(balanceContractAddr, cleanupEpochMethod, contract.All, epoch)
-
-	containerContractAddr := storage.Get(ctx, containerContractKey).(interop.Hash160)
-	contract.Call(containerContractAddr, cleanupEpochMethod, contract.All, epoch)
+	it := storage.Find(ctx, newEpochSubscribersKey, storage.RemovePrefix|storage.KeysOnly)
+	for iterator.Next(it) {
+		contractHash := interop.Hash160(iterator.Value(it).([]byte)[1:]) // one byte is for number prefix
```
AnnaShaleva, line 720: And here as well.

netmap/netmap_contract.go
```diff
@@ -601,6 +605,32 @@ func ListConfig() []record {
 	return config
 }
 
+// SubscribeForNewEpoch registers passed contract as a NewEpoch event
+// subscriber. Such a contract must have a `NewEpoch` method with a single
+// numeric argument. Transactions that call SubscribeForNewEpoch must be
+// witnessed by the Alphabet.
+// Produces `NewEpochHookRegistration` notification event with a just
+// registered recipient in a success case.
+func SubscribeForNewEpoch(contract interop.Hash160) {
```
AnnaShaleva, line 614: Just a question: don't we need unsubscription functionality?

netmap/netmap_contract.go
carpawell: https://github.com/nspcc-dev/neofs-contract/pull/368#discussion_r1402727722

netmap/netmap_contract.go
carpawell: not sure. that is my first contract _change_ (not extension) and that is how i would go but lets ask @roman-khimov

netmap/netmap_contract.go
carpawell: currently no. contracts should always be aware. mb for hotfixes only. in the future, i think, it is possible to extend (if needed) with a single update (a version that needs unsubscription also adds the unsubscribe method)

netmap/netmap_contract.go
roman-khimov: https://github.com/nspcc-dev/neofs-contract/blob/beb9293103656cef87475bc2532388ec2501e56f/neofsid/neofsid_contract.go#L56

Like this. It's absolutely critically important to keep these, even if we're not using them.

netmap/migration_test.go
```diff
@@ -24,17 +25,45 @@ func TestMigration(t *testing.T) {
...
+			if bytes.Equal(balanceHashOldKey, key) {
+				balanceHash, err = util.Uint160DecodeBytesLE(value)
+				require.NoError(t, err)
+
+				return
+			}
+		},
+	})
+
+	if containerHash.Equals(util.Uint160{}) || balanceHash.Equals(util.Uint160{}) {
```
cthulhu-rider, line 60:
```suggestion
	require.NotZerof(t, containerHash, "missing storage item %q with the Container contract address", containerHashOldKey)
	require.NotZerof(t, balanceHash, "missing storage item %q with the Balance contract address", balanceHashOldKey)
```

netmap/netmap_contract.go
```diff
@@ -73,7 +73,8 @@ const (
 	containerContractKey = "containerScriptHash"
 	balanceContractKey   = "balanceScriptHash"
 
-	cleanupEpochMethod = "newEpoch"
+	newEpochSubscribersKey = "newEpochSubscribers"
```
cthulhu-rider, line 76:
```suggestion
	newEpochSubscribersKeyPrefix = "newEpochSubscribers"
```

netmap/migration_test.go
```diff
@@ -184,3 +215,26 @@ func testMigrationFromDump(t *testing.T, d *dump.Reader) {
...
 }
+
+func checkNewEpochSubscribers(t *testing.T, contract *migration.Contract, balanceWant, containerWant util.Uint160) {
+	require.Nil(t, contract.GetStorageItem(balanceHashOldKey))
+	require.Nil(t, contract.GetStorageItem(containerHashOldKey))
+
+	// contracts are migrated in alphabetical order at least for now
+
+	contract.SeekStorage(append(newEpochSubsNewPrefix, 0), func(k, v []byte) bool {
+		balanceGot, err := util.Uint160DecodeBytesLE(k)
```
cthulhu-rider, line 226: u dont check func entrance. I'd gather map index->address and check after `SeekStorage` call

netmap/netmap_contract.go
```diff
@@ -601,6 +605,32 @@ func ListConfig() []record {
 	return config
 }
 
+// SubscribeForNewEpoch registers passed contract as a NewEpoch event
+// subscriber. Such a contract must have a `NewEpoch` method with a single
+// numeric argument. Transactions that call SubscribeForNewEpoch must be
```
cthulhu-rider, line 610:
```suggestion
// numeric parameter. Transactions that call SubscribeForNewEpoch must be
```

netmap/netmap_contract.go
```diff
@@ -601,6 +605,32 @@ func ListConfig() []record {
...
+// witnessed by the Alphabet.
+// Produces `NewEpochHookRegistration` notification event with a just
```
cthulhu-rider, line 612: according to code it is
```suggestion
// Produces `NewEpochSubscription` notification event with a just
```

tests/migration/storage.go
```diff
@@ -252,6 +252,13 @@ func (x *Contract) GetStorageItem(key []byte) []byte {
 	return x.exec.Chain.GetStorageItem(x.id, key)
 }
 
+// SeekStorage calls a provided handler against every stored item that starts
+// with a provided prefix. On handler's `false` return stops iteration. prefix
+// is removed from the resulting pair.
```
cthulhu-rider, line 257:
```suggestion
// is removed from the resulting pair key.
```
?

cthulhu-rider:
1st commit message has strange quotes

netmap/netmap_contract.go
roman-khimov: Please use one byte prefixes, they don't have to be this long and we can only have 64-byte keys.

netmap/netmap_contract.go
```diff
@@ -162,11 +161,19 @@ func _deploy(data any, isUpdate bool) {
 			switchToNotary(ctx)
 		}
 
-		return
-	}
+		if args.version <= 19_000 {
```
roman-khimov, line 164: `<`

common/netmap.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/nspcc-dev/neo-go/pkg/interop/contract"
+	"github.com/nspcc-dev/neo-go/pkg/interop/runtime"
+)
+
+// SubscribeForNewEpoch registers calling contract as a NewEpoch
+// callback requester. Netmap contract's address is taken from the
+// NNS contract, therefore, it must be presented and filled with
+// netmap information for a correct SubscribeForNewEpoch call; otherwise
+// a successive call is not guaranteed.
```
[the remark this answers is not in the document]
carpawell: > what call is this line talking about?

method call

> not guaranteed seems like some kind of random

i meant it depends on an NNS implementation. it may and may not panic

common/netmap.go
```diff
@@ -0,0 +1,17 @@
...
+// Caller must have `NewEpoch` method with a single numeric argument.
```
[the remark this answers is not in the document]
carpawell: @roman-khimov, no args' types check?

common/netmap.go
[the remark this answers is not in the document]
cthulhu-rider: i've read this doc several times and not sure i completely understand it. We impose some preconditions (Netmap contract registered in NNS). If complied with - success, otherwise panic, right? Where is `not guaranteed` come from?

`filled with netmap information` - who must be filled and what information?

common/netmap.go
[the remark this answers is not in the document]
roman-khimov: It doesn't care even for actual calls (hi, https://github.com/neo-project/neo/issues/1891) and you want it to check a complete signature in `hasMethod`...

netmap/netmap_contract.go
carpawell: returned

common/netmap.go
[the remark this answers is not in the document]
carpawell: > If complied with - success, otherwise panic, right?

well, to me that is not for sure. you need this to work it OK, but if it is not true, it depends on the NNS contract implementation. this call maybe a part of transaction, and panic (exception) may fail the whole TX

> Where is not guaranteed come from?

that is how i said "undefined behavior" but that time i wanted to say it this way

> filled with netmap information - who must be filled and what information?

NNS contract

common/netmap.go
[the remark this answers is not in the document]
carpawell: done

common/nns.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"github.com/nspcc-dev/neo-go/pkg/interop"
+	"github.com/nspcc-dev/neo-go/pkg/interop/contract"
+	"github.com/nspcc-dev/neo-go/pkg/interop/convert"
+	"github.com/nspcc-dev/neo-go/pkg/interop/lib/address"
+	"github.com/nspcc-dev/neo-go/pkg/interop/native/management"
+	"github.com/nspcc-dev/neo-go/pkg/interop/native/std"
+)
+
+// ResolveContractHash resolves contract hash by its well-known NeoFS name.
```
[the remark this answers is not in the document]
carpawell: > ResolveContractHash hash of the NeoFS contract 

what do you want to say by it?

> by its NNS domain name.

not sure we can say it. that is not the full NNS name, a caller should pass just `container`, not `container.neofs`

common/nns.go
```diff
@@ -0,0 +1,43 @@
...
+// Contract name should be lowercased, should not include `.neofs` TLD. Example
```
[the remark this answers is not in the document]
carpawell: > its better to either require name from the names list (rel thread above)

the list can grow and docs arent updated usually

> should be a valid NNS domain

but `container` is not a valid NNS name

common/nns.go
```diff
@@ -0,0 +1,43 @@
...
+// values: "netmap", "container", etc.
+// Relies on some NeoFS specifics:
+//  1. NNS contract should be deployed first (and should have `1` contract ID)
+//  2. It should be prefilled with contract hashes by their names (no
```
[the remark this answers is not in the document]
carpawell: NNS is kinda general thing and we are not the only ones who have it. so i wanted to highlight what exactly should be done: NNS is deployed and it should map `string` -> `uint160` (although i understand that the only callers of this method is are our contracts)

netmap/migration_test.go
carpawell: not expect it to fail ever (and if fail, it wont be a problem to understand the reason) but fixed

netmap/netmap_contract.go
carpawell: > newEpochSubscribersKeyPrefix

ok but without `key` then

> Please use one byte prefixes

`n` won my democratic election

netmap/migration_test.go
carpawell: fixed by bool flags

netmap/netmap_contract.go
carpawell: fixed

netmap/netmap_contract.go
carpawell: fixed. it did not live renamings well

tests/migration/storage.go
carpawell: fixed

netmap/netmap_contract.go
carpawell: fixed

carpawell:
> 1st commit message has strange quotes

fixed

common/netmap.go
```diff
@@ -12,6 +14,11 @@ import (
 // a successive call is not guaranteed.
 // Caller must have `NewEpoch` method with a single numeric argument.
 func SubscribeForNewEpoch() {
+	callingHash := runtime.GetExecutingScriptHash()
+	if !management.HasMethod(callingHash, "newEpoch", 1) {
+		panic(std.Base58Encode(callingHash) + " contract does not have `newEpoch(epoch)` method")
```
roman-khimov, line 19: `std.Base58Encode` is not the best choice here, it's not an address and it's not a hash representation. You can try https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/interop@v0.0.0-20231123162724-195bb742ca16/native/std#Itoa or https://github.com/nspcc-dev/neo-go/blob/195bb742ca16/pkg/interop/lib/address/address.go#L28.

carpawell: ok, `address.FromHash160`

common/netmap.go
```diff
@@ -12,6 +14,11 @@ import (
 // a successive call is not guaranteed.
 // Caller must have `NewEpoch` method with a single numeric argument.
 func SubscribeForNewEpoch() {
+	callingHash := runtime.GetExecutingScriptHash()
```
roman-khimov, line 17: It's executing script hash and you're calling it `calling`.

common/netmap.go
```diff
@@ -12,6 +14,11 @@ import (
...
+	if !management.HasMethod(callingHash, "newEpoch", 1) {
```
roman-khimov, line 18: Why this check is here and not in the `subscribeForNewEpoch` itself? It makes much more sense there since `netmap` controls its state and shouldn't allow bad subscriptions (forget contract upgrades for a moment).

common/netmap.go
carpawell: i meant that is a contract that called that method. renamed to `callingContract`, it is ok?

common/nns.go
```diff
@@ -0,0 +1,43 @@
...
+//     capitalized chars; no `neofs` TLD)
+func ResolveContractHash(contractName string) interop.Hash160 {
+	// get NNS contract (it always has ID=1 in the NeoFS Sidechain)
+	nnsContract := management.GetContractByID(1)
+	if nnsContract == nil {
+		panic("missing NNS contract")
+	}
+
+	resResolve := contract.Call(nnsContract.Hash, "resolve", contract.ReadOnly,
+		contractName+".neofs", 16, // TXT
```
AnnaShaleva, line 27: How about `nns.TXT`?

AnnaShaleva (changes requested):
~@carpawell, please, pay attention to https://github.com/nspcc-dev/neofs-contract/pull/368/commits/d2201e4fbd360fc77e8e3807071e4e5ac2532724#r1403628959.~

Good, the notifications ordering is needed. Otherwise LGTM.

netmap/netmap_contract.go
```diff
@@ -73,7 +74,8 @@ const (
 	containerContractKey = "containerScriptHash"
```
AnnaShaleva, line 74: We may add `nolint` comment, we need to keep our linter job with OK status.

carpawell: oh, i think i got it. it is used only in `_deploy` func but this func is nolinted already and is ignored by the linter and the linter does not see its usage

added `nolint`

common/nns.go
carpawell: import cycle, unfortunately. that is how i found https://github.com/nspcc-dev/neo-go/issues/3202

netmap/netmap_contract.go
```diff
@@ -601,6 +608,37 @@ func ListConfig() []record {
...
+	for iterator.Next(it) {
+		num += 1
+	}
+
+	var key []byte
+	key = append(key, []byte(newEpochSubscribersPrefix)...)
+	key = append(key, num)
+	key = append(key, contract...)
+
+	storage.Put(ctx, key, []byte{})
```
AnnaShaleva, line 639: @roman-khimov, wrt https://github.com/nspcc-dev/neofs-contract/pull/368#discussion_r1403628376 and https://github.com/nspcc-dev/neofs-contract/pull/368#discussion_r1402727722: we need ordering guarantees and it's good. But this scheme currently allows subscriptions duplicates I think that we need to disallow subscriptions duplicates on the netmap contract level, and in this part of code we need to add a check whether subscription is already added.

What do you think?

netmap/netmap_contract.go
```diff
@@ -601,6 +610,37 @@ func ListConfig() []record {
...
+// SubscribeForNewEpoch registers passed contract as a NewEpoch event
+// subscriber. Such a contract must have a `NewEpoch` method with a single
+// numeric parameter. Transactions that call SubscribeForNewEpoch must be
+// witnessed by the Alphabet.
+// Produces `NewEpochSubscription` notification event with a just
+// registered recipient in a success case.
+func SubscribeForNewEpoch(contract interop.Hash160) {
+	common.CheckAlphabetWitness(common.AlphabetAddress())
+
+	callingContract := runtime.GetExecutingScriptHash()
```
roman-khimov, line 622: And here you have `contract`, executing script hash is netmap.

carpawell: wow, fixed

netmap/netmap_contract.go
```diff
@@ -73,7 +74,8 @@ const (
 	containerContractKey = "containerScriptHash"
 	balanceContractKey   = "balanceScriptHash"
 
-	cleanupEpochMethod = "newEpoch"
+	newEpochSubscribersPrefix = "n"
```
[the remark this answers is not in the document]
carpawell: ok, now that is `e`

netmap/netmap_contract.go
carpawell: done with ignoring (no logs or panics)

netmap/netmap_contract.go
```diff
@@ -601,6 +610,40 @@ func ListConfig() []record {
...
+
+	if !management.HasMethod(contract, "newEpoch", 1) {
+		panic(address.FromHash160(contract) + " contract does not have `newEpoch(epoch)` method")
+	}
+
+	ctx := storage.GetContext()
+	var num byte
+	it := storage.Find(ctx, []byte(newEpochSubscribersPrefix), storage.KeysOnly|storage.RemovePrefix)
+	for iterator.Next(it) {
+		if common.BytesEqual(iterator.Value(it).(interop.Hash160), contract) {
```
roman-khimov, line 630: So, you're searching through `newEpochSubscribersPrefix`, removing prefix and they what you end up with is 21 bytes, index and hash. You need to strip index at least and then maybe use https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/interop#Hash160.Equals

carpawell: > you end up with is 21 bytes

my bad, fixed

> maybe use https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/interop#Hash160.Equals

ok, looks better

roman-khimov (dismissed):

netmap/netmap_contract.go
```diff
@@ -685,9 +717,9 @@ func setConfig(ctx storage.Context, key, val any) {
...
 func cleanup(ctx storage.Context, epoch int) {
-	balanceContractAddr := storage.Get(ctx, balanceContractKey).(interop.Hash160)
-	contract.Call(balanceContractAddr, cleanupEpochMethod, contract.All, epoch)
-
-	containerContractAddr := storage.Get(ctx, containerContractKey).(interop.Hash160)
-	contract.Call(containerContractAddr, cleanupEpochMethod, contract.All, epoch)
+	it := storage.Find(ctx, newEpochSubscribersPrefix, storage.RemovePrefix|storage.KeysOnly)
+	for iterator.Next(it) {
+		contractHash := interop.Hash160(iterator.Value(it).([]byte)[1:]) // one byte is for number prefix
+		contract.Call(contractHash, cleanupEpochMethod, contract.All, epoch)
```
AnnaShaleva, line 735: And one more important moment that just came to my mind. If something (panic, I mean) happens in the calling callback, then the rest of callbacks won't be executed. It remains `PostPersist` and technically it's OK to fail immediately, so just a question: do we need to think about some exception handlers here?

netmap/netmap_contract.go
```diff
@@ -623,8 +623,13 @@ func SubscribeForNewEpoch(contract interop.Hash160) {
 
 	ctx := storage.GetContext()
 	var num byte
-	it := storage.Find(ctx, []byte(newEpochSubscribersPrefix), storage.None)
+	it := storage.Find(ctx, []byte(newEpochSubscribersPrefix), storage.KeysOnly|storage.RemovePrefix)
 	for iterator.Next(it) {
+		raw := iterator.Value(it).([]byte)[1:] // 1 byte is an index
+		if contract.Equals(raw) {
+			return
+		}
```
AnnaShaleva, line 633: I'm a paranoiac about VM's `EQUAL` and compiler, could we have a test for the contract that subscribes for notifications and after that subscribes one more time? In the end we should check that there's only one subscriber record stored in the netmap.

AnnaShaleva (approved):
LGTM otherwise, but these are minor comments and may be left unresolved.

netmap/netmap_contract.go
```diff
@@ -95,8 +99,8 @@ func _deploy(data any, isUpdate bool) {
 
 	var args = data.(struct {
 		notaryDisabled bool
-		addrBalance    interop.Hash160
-		addrContainer  interop.Hash160
+		addrBalance    interop.Hash160 // Balance contract not used legacy
+		addrContainer  interop.Hash160 // Container contract not used legacy
```
AnnaShaleva, line 103: BTW, if these two fields are unused, then you can explicitly mark them as `_`.

netmap/netmap_contract.go
carpawell: > do we need to think about some exception handlers here?

only our own contracts are expected to be new-epoch-notified so any panic is a global panic to me and i can not say that some contract is more or less important to us, we need all of them to be notified. failed notifications are a failed NeoFS network to me. also, i do know what such a handler should do. notify, panic but later, log?

@roman-khimov

netmap/netmap_contract.go
carpawell: fixed

netmap/netmap_contract.go
roman-khimov: Partial failure in many ways is better than complete failure, so handling exceptions makes some sense here, but I'd leave it to another issue.

roman-khimov (dismissed):

carpawell:
Some var renamings, commit body fixes, sry.

PR #379, opened 2023-11-29
nns: fix deployments with 0.38.1 adm

roman-khimov:
I'm not yet sure we need it, but old adm (before nspcc-dev/neofs-node#2625) creates a broken deploy script with actual data that has no fields inside. Not a problem for updates, not a problem for newer adm, but still.


carpawell (approved):

carpawell:
So what do we do?

roman-khimov:
0.38.1 can't be fixed. So maybe this patch can be thrown away as irrelevant for real life.

carpawell:
> this patch can be thrown away

I think we can make it.

PR #389, opened 2024-02-26
Go 1.20

roman-khimov:


carpawell (dismissed):

.github/workflows/go.yml
```diff
@@ -60,7 +60,6 @@ jobs:
         run: make build
 
       - name: Check that fresh version of RPC bindings is committed (for minimum supported Go version only)
-        if: ${{ matrix.go_versions == '1.19' }}
```
AnnaShaleva, line 63: It's not version-dependent, I agree, but the idea was to check it only once, we don't need to perform it three times in a raw.

But if you'd like to, then comment should also be adjusted: `(for minimum supported Go version only)`.

AnnaShaleva: And BTW, I think that in some very special cases it may be version-dependent due to https://github.com/nspcc-dev/neo-go/issues/3133.

roman-khimov: The cost of this check is negligible compared to `make build` itself, so. I'll fix the comment.

.github/workflows/go.yml
```diff
@@ -10,10 +10,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.19
-      - uses: actions/checkout@v3
+          go-version-file: 'go.mod'
```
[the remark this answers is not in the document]
roman-khimov: We're rolling it throughout our repos for some time already starting with neofs-node.

contracts/alphabet/migration_test.go
```diff
@@ -32,7 +32,7 @@ func replaceArgI(vs []any, i int, v any) []any {
 }
 
 func randUint160() (u util.Uint160) {
-	rand.Read(u[:])
+	rand.Read(u[:]) //nolint:staticcheck // SA1019: rand.Read has been deprecated since Go 1.20
```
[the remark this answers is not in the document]
roman-khimov: `crypto/rand` costs a bit more. Yeah, probably not a problem for tests either, but this code works already, there is no point in changing it.

AnnaShaleva (approved):

.github/workflows/go.yml
AnnaShaleva: And still, you do this three times instead of one. And have three identical (failing or successful) test reports instead of one report. So this doesn't look good to me.

AnnaShaleva: Although it's not a separate job, this check is bound to the building job, so OK.

carpawell (approved):

PR #396, opened 2024-04-21
workflows: test on MacOS (arm64) and Windows

roman-khimov:


.github/workflows/go.yml
```diff
@@ -22,18 +22,29 @@ jobs:
 
   tests:
     name: Tests
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        go_versions: [ '1.20', '1.21', '1.22' ]
+        go: [ '1.20', '1.21', '1.22' ]
+        os: [ubuntu-latest, windows-2022, macos-14]
```
carpawell, line 29: why some are the latest while the others are fixed?

carpawell (approved):

.github/workflows/go.yml
roman-khimov: IIRC `latest` there is not what we have here.

PR #397, opened 2024-05-02
NNS wrapper extensions

roman-khimov:
Inspired by https://github.com/nspcc-dev/neofs-node/pull/2841.

### DIFF
--- a/rpc/nns/hashes.go
+++ b/rpc/nns/hashes.go
@@ -36,27 +36,50 @@ func InferHash(sg ContractStateGetter) (util.Uint160, error) {
 	return c.Hash, nil
 }
 
+// AddressFromRecord extracts [util.Uint160] hash from the string using one of
+// the following formats:
+//   - hex-encoded LE (reversed) string
+//   - Neo address ("Nxxxx")
+//
+// NeoFS used both for contract hashes stored in NNS at various stages of its
+// development.
+//
+// See also: [AddressFromRecords].
+func AddressFromRecord(s string) (util.Uint160, error) {
+	h, err := util.Uint160DecodeStringLE(s)
+	if err == nil {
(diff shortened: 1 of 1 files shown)


rpc/nns/hashes.go
```diff
@@ -36,27 +36,69 @@ func InferHash(sg ContractStateGetter) (util.Uint160, error) {
...
+func NewFSReader(sg ContractStateGetter, invoker Invoker) (*ContractReader, error) {
+	h, err := InferHash(sg)
+	if err != nil {
+		return nil, err
+	}
+	return NewReader(invoker, h), nil
+}
+
+// NewFS creates an instance of [Contract] using hash obtained via [InferHash].
+func NewFS(sg ContractStateGetter, actor Actor) (*Contract, error) {
```
cthulhu-rider, line 50: confusing name to me, i suggest
```suggestion
func NewInferred(sg ContractStateGetter, actor Actor) (*Contract, error) {
```
or
```suggestion
func NewWithInferredHash(sg ContractStateGetter, actor Actor) (*Contract, error) {
```

cthulhu-rider (commented):
nice except some naming

rpc/nns/hashes.go
roman-khimov: Yeah, I had some doubts about it as well, but decided to try with a shorter one. @carpawell, what'd be your preference?

carpawell: yes, i would die if i did not know what and why FS is here. NewInferred and NewWithInferredHash are not perfect to me either but better

roman-khimov: Fixed.

carpawell (approved):

PR #401, opened 2024-06-14
Contract binary package

roman-khimov:


scripts/compare-fscontent/go.mod
```diff
@@ -0,0 +1,34 @@
+module github.com/nspcc-dev/neo-go/scripts/compare-fscontent
```
cthulhu-rider, line 1:
```suggestion
module github.com/nspcc-dev/neofs-contract/scripts/compare-fscontent
```
?

deploy/nns.go
```diff
@@ -190,7 +191,7 @@ func lookupNNSDomainRecord(inv *invoker.Invoker, nnsContract util.Uint160, domai
 	item, err := unwrap.Item(inv.Call(nnsContract, methodNNSResolve, domainName, int64(nns.TXT)))
 	if err != nil {
 		var except unwrap.Exception
-		if errors.As(err, &except) && string(except) == "token not found" {
+		if errors.As(err, &except) && strings.Contains(string(except), "token not found") {
```
cthulhu-rider, line 194: became less laconic at this scope to me. Strange to see `invocation failed: at instruction 4100 (THROW): unhandled exception: \"token not found\""` as `unwrap.Exception` when (Go) contract just makes `panic("token not found")`. Is it possible to improve somehow to keep previous code?

contracts/contracts_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/manifest"
+	"github.com/nspcc-dev/neo-go/pkg/smartcontract/nef"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetFS(t *testing.T) {
+	c, err := GetFS()
+	require.NoError(t, err)
+	require.Equal(t, len(fsContracts), len(c))
```
cthulhu-rider, line 16: suggest to also test `They're returned in the order they're supposed to be deployed starting
from NNS` statement from docs. We can also check exact (i guess) result contents since embedded files are now presented in the repo

ok to be done later

contracts/contracts.go
```diff
@@ -0,0 +1,125 @@
...
+		reputationDir,
+	}
+	mainContracts = []string{
+		neofsDir,
+		processingDir,
+	}
+)
+
+// GetFS returns current set of NeoFS chain contracts stored in the package.
+// They're returned in the order they're supposed to be deployed starting
```
cthulhu-rider, line 67: but the order (`fsContracts`) differs from the implemented deployment procedure, is it intentional?

cthulhu-rider (commented):
nice overall

deploy/nns.go
roman-khimov: It's not, RPC server returns `at instruction 4100 (THROW): unhandled exception: "token not found"`. Exception here means `exception` field in the RPC protocol which is a broader thing than contract's exception (ABORT will produce a similar one, for example, but it's not an exception at all).

contracts/contracts.go
roman-khimov: Fixed.

scripts/compare-fscontent/go.mod
roman-khimov: Fixed.

cthulhu-rider (approved):

PR #405, opened 2024-06-18
Micro-polishing

### LINKED ISSUE #307: Implement some of good practices for the contracts
Some of our contracts contain issues that need to be refactored:
1. Replace `common.BytesEqual` usages with built-in `(a Hash160) Equals(b interface{}) bool` where possible (see https://github.com/nspcc-dev/neofs-contract/blob/dc89137484cd042e79378d4745558cc560b6368a/alphabet/alphabet_contract.go#L30 as an example). The same thing can be done for Hash256 comparison cases.
2. Use pre-defined constants where appropriate (`interop.Hash160Len`, `interop.PublicKeyCompressedLen`, etc.). There are still several cases where numbers are use instead (see https://github.com/nspcc-dev/neofs-contract/blob/dc89137484cd042e79378d4745558cc560b6368a/container/container_contract.go#L472 as an example).
3. Use the most specific parameter types for method signatures where appropriate (e.g. `interop.Hash160` instead of `[]byte`), see https://github.com/nspcc-dev/neofs-contract/blob/dc89137484cd042e79378d4745558cc560b6368a/neofs/neofs_contract.go#L371 as an example.

Ref. #304.

roman-khimov:
Fixes #307.


contracts/neofs/contract.go
```diff
@@ -369,7 +369,7 @@ func Cheque(id []byte, user interop.Hash160, amount int, lockAcc []byte) {
 //
 // This method produces Bind notification. This method panics if keys are not
 // 33 byte long. User argument must be a valid 20 byte script hash.
```
cthulhu-rider, line 371: now this requirement follows from the type, mb drop?

cthulhu-rider (approved):

contracts/neofs/contract.go
roman-khimov: Well, keys are type-specific for a long time as well and yet there is something about their length above. This doesn't make much sense to me, but doesn't really hurt either.

PR #406, opened 2024-06-20
Add prefixes to balance contract

roman-khimov:
And some tests as well.


contracts/balance/migration_test.go
```diff
@@ -60,7 +62,25 @@ func testMigrationFromDump(t *testing.T, d *dump.Reader) {
 	// try to update the contract
 	if notaryDisabled && prevPendingVotes {
 		c.CheckUpdateFail(t, "pending vote detected")
-		return
+		t.Fail()
```
cthulhu-rider, line 65: have u reached this? the change worth extension to tests of other contracts (and separation from the currentone)

contracts/balance/migration_test.go
```diff
@@ -60,7 +62,25 @@ func testMigrationFromDump(t *testing.T, d *dump.Reader) {
...
+	}
+
+	var accounts = []util.Uint160{}
```
cthulhu-rider, line 68: cosmetic
```suggestion
	var accounts []util.Uint160
```
same below

cthulhu-rider, line 68: i'd also assert that items with these keys are removed from the storage after update

contracts/balance/migration_test.go
roman-khimov: No. We don't have it in the current data set and we won't ever have it for obvious reasons, but I wanted to ensure that test doesn't silently "succeed". Can be reverted.

cthulhu-rider (approved):

PR #410, opened 2024-06-21
deploy: transfer GAS in the first place if notaries are set

### LINKED ISSUE #409: Speed up NeoFS chain deployment from scratch with `set_roles_in_genesis` option
IR launched in auto-deployment mode always deploys NNS first when starting from scratch. This requires some time to wait blocks so so that the deployer (now it's always committee member Az) has enough GAS

when  option
https://github.com/nspcc-dev/neofs-node/blob/5c66479ee4bbd727c74abab8e635eb74aea4cf9d/config/example/ir.yaml#L86-L88
is enabled, IR is able to fill committee accounts with GAS insta cuz Notary system is completely ready for work starting from the genesis block 

in auto-deployment procedure, 1st of all check the Notary service's readiness for consumption by the committee. If roles are unset, continue as always (deploy NNS and so on), otherwise, make GAS transfer stage 1st, and then perform all other stages with ready-to-go GAS. Taking into account how much GAS is required for the NNS deployment, this approach should speed up the whole procedure for deploying the network

leave anything untouched: everything works now but takes a bit more time (depending on
https://github.com/nspcc-dev/neofs-node/blob/5c66479ee4bbd727c74abab8e635eb74aea4cf9d/config/example/ir.yaml#L37-L38
to be more precise). Need to see how much we win with proposed solution

carpawell:
It is possible to start blockchain with required roles already set. In this case, there is no need to wait for NNS to be deployed for GAS transfers, since notary service (requires Notary roles to be set correctly) is ready. It allows all multi-signed operations to be applied at an earlier stage. Closes #409.

### DIFF
--- a/deploy/alphabet.go
+++ b/deploy/alphabet.go
@@ -55,22 +55,13 @@ func designateNeoFSAlphabet(ctx context.Context, prm initAlphabetPrm) error {
 
 		prm.logger.Info("checking NeoFS Alphabet role of the committee members...")
 
-		accsWithAlphabetRole, err := roleContract.GetDesignatedByRole(noderoles.NeoFSAlphabet, prm.monitor.currentHeight())
+		ok, err := checkRole(noderoles.NeoFSAlphabet, &roleContract.ContractReader, prm.monitor, prm.committee)
 		if err != nil {
-			prm.logger.Error("failed to check role of the committee, will try again later", zap.Error(err))
+			prm.logger.Error("failed to check NeoFS Alphabet role of the committee, will try again later", zap.Error(err))
 			continue
 		}
 
-		someoneWithoutRole := len(accsWithAlphabetRole) < len(prm.committee)
-		if !someoneWithoutRole {
-			for i := range prm.committee {
-				if !accsWithAlphabetRole.Contains(prm.committee[i]) {
-					someoneWithoutRole = true
-					break
-				}
-			}
-		}
-		if !someoneWithoutRole {
+		if ok {
 			prm.logger.Info("all committee members have a NeoFS Alphabet role")
 			return nil
 		}
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -225,12 +225,12 @@ func Deploy(ctx context.Context, prm Prm) error {
 		return fmt.Errorf("init blockchain monitor: %w", err)
 	}
 
-	rolesAreSet, err := checkCommitteeRoles(prm.Logger, prm.Blockchain, monitor, committee)
+	notaryRolesSet, alphaRolesSet, err := checkCommitteeRoles(prm.Blockchain, monitor, committee)
 	if err != nil {
 		return fmt.Errorf("pre-check committee roles: %w", err)
 	}
-	transfersDone := rolesAreSet // currently the same
-	if rolesAreSet {
+	transfersDone := notaryRolesSet // currently the same
+	if notaryRolesSet {
 		// if it is possible to transfer GAS before any network settings,
 		// it should be done as it speeds up expensive operations
 
@@ -277,7 +277,7 @@ func Deploy(ctx context.Context, prm Prm) error {
 
 	prm.Logger.Info("NNS contract successfully initialized on the chain", zap.Stringer("address", nnsOnChainAddress))
 
-	if !rolesAreSet {
+	if !notaryRolesSet {
 		prm.Logger.Info("enable Notary service for the committee...")
 
 		err = enableNotary(ctx, enableNotaryPrm{
@@ -318,7 +318,7 @@ func Deploy(ctx context.Context, prm Prm) error {
 		prm.Logger.Info("initial transfer to the committee successfully done")
 	}
 
-	if !rolesAreSet {
+	if !alphaRolesSet {
 		prm.Logger.Info("initializing NeoFS Alphabet...")
 
 		err = designateNeoFSAlphabet(ctx, initAlphabetPrm{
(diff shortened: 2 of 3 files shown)


deploy/deploy.go
```diff
@@ -645,3 +684,52 @@ func neoFSRuntimeTransactionModifier(getBlockchainHeight func() uint32) actor.Tr
 		return nil
 	}
 }
+
+func rolesAreSet(b Blockchain, committee keys.PublicKeys, logger *zap.Logger) (bool, error) {
```
cthulhu-rider, line 688: this essentially is a combination of same checks done in `enableNotary` and `designateNeoFSAlphabet`, i suggest to share and reuse the code

deploy/deploy.go
```diff
@@ -205,11 +208,50 @@ func Deploy(ctx context.Context, prm Prm) error {
...
+	go autoReplenishNotaryBalance(ctx, prm.Logger, prm.Blockchain, prm.LocalAccount, chNewBlock)
+
 	monitor, err := newBlockchainMonitor(prm.Logger, prm.Blockchain, chNewBlock)
 	if err != nil {
 		return fmt.Errorf("init blockchain monitor: %w", err)
 	}
 
+	rolesSet, err := rolesAreSet(prm.Blockchain, committee, prm.Logger)
+	if err != nil {
+		return fmt.Errorf("roles check: %w", err)
```
cthulhu-rider, line 230:
```suggestion
		return fmt.Errorf("pre-check committee roles: %w", err)
```

deploy/deploy.go
cthulhu-rider, line 688: calling such function a statement is strange, for me it’s `checkCommitteeRoles`

deploy/deploy.go
```diff
@@ -645,3 +684,52 @@ func neoFSRuntimeTransactionModifier(getBlockchainHeight func() uint32) actor.Tr
 		return nil
 	}
 }
+
+func rolesAreSet(b Blockchain, committee keys.PublicKeys, logger *zap.Logger) (bool, error) {
+	currHeight, err := b.GetBlockCount()
+	if err != nil {
+		return false, fmt.Errorf("current block height: %w", err)
```
cthulhu-rider, line 691: u can just accept `monitor.currentHeight()`

cthulhu-rider (commented):
correct overall, left several cosmetic suggestions

deploy/deploy.go
carpawell: they do almost the same things now, do you think `enableNotary` and `designateNeoFSAlphabet` need this check after this PR?

deploy/deploy.go
carpawell: i need monitor here then (one more param) but ok, fixed

deploy/deploy.go
cthulhu-rider: ofc they need: each time (block) they attempt to set roles - they need to check if it worked or not

to be more precise, my discussion is not only about the code sharing but diff behavior. IIUC 
1. `rolesAsSet` checks exact equality of committee and role holders
2. `enableNotary` and `designateNeoFSAlphabet` just check committee occurrence by subset (ignoring situation when there are some role members besides the committee)

so, code sharing would prevent such diff

lets revise desired behavior. `2` is weaker, and deploy mechanics should work well in this case (never tested tbh). Can it cause any problems idk about? @roman-khimov @AnnaShaleva

carpawell: > > do you think enableNotary and designateNeoFSAlphabet need this check after this PR?
>
> ofc they need

with this PR, roles are either set or not, `enableNotary` can be called if Notary role is not set and can be not called at all otherwise

> each time (block) they attempt to set roles - they need to check if it worked or not

this code surprised me. my practice says that a correct TX acceptance is enough, otherwise our life is too paranoid here. `neofs-adm` does not do such additional checks, as i remember

> enableNotary and designateNeoFSAlphabet just check committee occurrence by subset 

do we really expect and are ready for committee < Notary/Alphabet case?

cthulhu-rider: adm has direct access to all committee and consensus accounts, ofc its behavior is simpler. Here only 1 account acts, i dont think logic can be easier than now

carpawell: > i dont think logic can be easier than now

we have TX, we have VUB, why do we need a polling mechanism?

i do not say it is already done (or should be done) in this PR, i just want to make clear what exact code you want to share. currently, it seems to me like a wrong assumption that setting a role should be checked with getting a role (every block). if we agree that it is bad, i do not think sharing is worth it then, we just need to create an issue and this part will be removed

roman-khimov: > do we really expect and are ready for committee < Notary/Alphabet case?

`committee < alphabet` is impossible to me, maybe unless you play with CommitteeHistory/ValidatorsHistory in some bad way, but even then validators are always a subset of committee and alphabet nodes are always validators. Committee and Notary role relations allow for a bit more freedom, one _can_ technically have more notary nodes than we have them in committee, but I don't know any real use case for this. Either way, current code _always_ sets both to the committee (kicking everyone else out), so on a properly configured system we can expect them to be in a perfect sync. Therefore, we can check for exact match everywhere.

> we have TX, we have VUB, why do we need a polling mechanism?

You have to probe for configuration (read/compare lists) anyway to see if it needs to be changed. This covers for transaction execution side-effects perfectly which are more reliable than just waiting for successful transaction execution. Can't say that just waiting for the transaction outcome is wrong, it works as well (you know how contracts you're calling work, they will fail the transaction in case anything is wrong) and it's a bit more efficient (subscriptions!), but since you already have some code that reads configurations reusing it isn't _that_ bad either.

carpawell: i do not understand. i have shared the code (that has not been shared b/w roles itself before this PR too) that i think should be different

cthulhu-rider: transaction is just a mechanism, routine wait for particular state, it does not care how this state was achieved. This covers two additional cases:
* if an earlier transaction has already been completed, there is no need to wait for this one, it will be no-op anyway
* if the transaction is successful, but the state still does not satisfying

so I see no reason to do different pre-checks and post-checks, the state condition should be checked uniformly

deploy/alphabet.go
```diff
@@ -55,22 +55,12 @@ func designateNeoFSAlphabet(ctx context.Context, prm initAlphabetPrm) error {
 
 		prm.logger.Info("checking NeoFS Alphabet role of the committee members...")
 
-		accsWithAlphabetRole, err := roleContract.GetDesignatedByRole(noderoles.NeoFSAlphabet, prm.monitor.currentHeight())
+		ok, err := checkRole(noderoles.NeoFSAlphabet, prm.blockchain, prm.monitor, prm.committee)
 		if err != nil {
-			prm.logger.Error("failed to check role of the committee, will try again later", zap.Error(err))
-			continue
+			return fmt.Errorf("role check: %w", err)
```
cthulhu-rider, line 60: do not change behavior: log and `continue`, not return. Same in `enableNotary`

deploy/notary.go
```diff
@@ -84,31 +84,19 @@ func enableNotary(ctx context.Context, prm enableNotaryPrm) error {
 		}
 	}
 
-	roleContract := rolemgmt.NewReader(invoker.New(prm.blockchain, nil))
```
cthulhu-rider, line 87: decided to init it again and again?

cthulhu-rider (changes requested):

deploy/notary.go
carpawell: it only does `return &ContractReader{invoker}` inside but ok

deploy/alphabet.go
carpawell: i just deleted this part and wrote from scratch so this is why i missed it. overall, it is strange to me: only caller's context can stop this then, all the errors are just ignored (how do we expect to check rolles, if no network is available?). just a user's experience issue to me. however, i am for keeping behavior in commits that do not declare changes, fixed

roman-khimov (dismissed):

deploy/deploy.go
```diff
@@ -645,3 +684,39 @@ func neoFSRuntimeTransactionModifier(getBlockchainHeight func() uint32) actor.Tr
...
+		return false, false, fmt.Errorf("%s role check: %w", noderoles.NeoFSAlphabet, err)
+	}
+
+	return notaryRole, alphaRole, nil
+}
+
+func checkRole(role noderoles.Role, roleContract *rolemgmt.ContractReader, m *blockchainMonitor, committee keys.PublicKeys) (bool, error) {
+	currentRoles, err := roleContract.GetDesignatedByRole(role, m.currentHeight())
+	if err != nil {
+		return false, fmt.Errorf("reading %s role: %w", role, err)
```
cthulhu-rider, line 708: ahhh it always kills me when error context contains function input, caller will most likely get tautology

deploy/notary.go
```diff
@@ -93,22 +93,13 @@ func enableNotary(ctx context.Context, prm enableNotaryPrm) error {
 
 		prm.logger.Info("checking Notary role of the committee members...")
 
-		accsWithNotaryRole, err := roleContract.GetDesignatedByRole(noderoles.P2PNotary, prm.monitor.currentHeight())
+		ok, err := checkRole(noderoles.P2PNotary, roleContract, prm.monitor, prm.committee)
 		if err != nil {
-			prm.logger.Error("failed to check role of the committee, will try again later", zap.Error(err))
+			prm.logger.Error("failed to check notary role of the committee, will try again later", zap.Error(err))
```
cthulhu-rider, line 98:
```suggestion
			prm.logger.Error("failed to check Notary role of the committee, will try again later", zap.Error(err))
```
to follow other messages

deploy/alphabet.go
```diff
@@ -55,22 +55,13 @@ func designateNeoFSAlphabet(ctx context.Context, prm initAlphabetPrm) error {
 
 		prm.logger.Info("checking NeoFS Alphabet role of the committee members...")
 
-		accsWithAlphabetRole, err := roleContract.GetDesignatedByRole(noderoles.NeoFSAlphabet, prm.monitor.currentHeight())
+		ok, err := checkRole(noderoles.NeoFSAlphabet, &roleContract.ContractReader, prm.monitor, prm.committee)
 		if err != nil {
-			prm.logger.Error("failed to check role of the committee, will try again later", zap.Error(err))
+			prm.logger.Error("failed to check alphabet role of the committee, will try again later", zap.Error(err))
```
cthulhu-rider, line 60:
```suggestion
			prm.logger.Error("failed to check NeoFS Alphabet role of the committee, will try again later", zap.Error(err))
```

cthulhu-rider (approved):
ok now, some cosmetics left

deploy/deploy.go
carpawell: overall agree

cthulhu-rider (approved):

PR #417, opened 2024-07-11
deploy: copy transaction expected to be changed

carpawell:
The whole `makeUnsignedDesignateCommitteeNotaryTx` is designed to be called many times if something goes wrong. But currently neo-go's actor shares signers with TXs it produces, so the actor is broken immediately after this func is called. Can be seen only if transaction-to-sign is expired, in practice happened when deployment nodes were started with a huge delay.

### DIFF
--- a/deploy/notary.go
+++ b/deploy/notary.go
@@ -831,6 +831,8 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
 		return nil, err
 	}
 
+	tx = tx.Copy()
+
 	tx.ValidUntilBlock = sharedTxData.validUntilBlock
 	tx.Nonce = sharedTxData.nonce
 	tx.Signers[0].Account = sharedTxData.sender


deploy/notary.go
```diff
@@ -831,6 +831,8 @@ func makeUnsignedDesignateCommitteeNotaryTx(roleContract *rolemgmt.Contract, com
 		return nil, err
 	}
 
+	tx = tx.Copy()
```
cthulhu-rider, line 834: enough to copy and modify `Signers` if `!tx.Signers[0].Account.Equals(sharedTxData.sender)` only

cthulhu-rider:
shared memory is not the root of the problem. According to the actor concept, the resulting tx should not be modified due to calculated fees. Therefore, we should use a special actor instead of modifying the tx

the closest approach could be [`TransactionModifier`](https://pkg.go.dev/github.com/nspcc-dev/neo-go@v0.106.2/pkg/rpcclient/actor#Options) opt, but
> MakeUnsigned* methods do not run it.

lets create an issue in Neo Go project to find the best solution

deploy/notary.go
carpawell: this is kinda hotfix to me, i do not agree with the code here, so something different can be done here in the future. do you really need one more `if` here?

carpawell: btw, this func is called only if `!sharedTxDataMatches` so suggested `if` is already done

cthulhu-rider: >  i do not agree with the code here, so something different can be done here in the future

then issue or to draft

cthulhu-rider:
> How should it sound?

i'll do my best to create it

deploy/notary.go
cthulhu-rider: > this is kinda hotfix to me

nvm, we dont modify other ref type fields, so copy only `Signers`

carpawell:
> we can merge it only with issue+FIXME

https://github.com/nspcc-dev/neofs-contract/issues/418. I want to be sure any bigger change works and want to see if there are more places with similar problems. Currently, I have to concentrate on deployment and just want a fix in the master.

cthulhu-rider (approved):
https://github.com/nspcc-dev/neofs-contract/pull/417/files#r1676133208

roman-khimov (approved):

PR #419, opened 2024-07-29
NNS update for 0.20.0

roman-khimov:
It's a rebase of https://github.com/nspcc-dev/neofs-contract/pull/266 with two additional commits.


contracts/nns/contract.go
```diff
@@ -514,7 +514,12 @@ func SetRecord(name string, typ recordtype.Type, id byte, data string) {
 	ctx := storage.GetContext()
 	ns := getFragmentedNameState(ctx, tokenID, fragments)
 	ns.checkAdmin()
-	putRecord(ctx, tokenID, name, typ, id, data)
+	recordKey := getIdRecordKey(tokenID, name, typ, id)
```
carpawell, line 515: but why? both ways are ok to me but cannot agree with "it makes the contract code more complicated"
making a small description of a part of the code (making it a named func call) is not a complication to me

contracts/nns/contract.go
```diff
@@ -60,7 +60,7 @@ const (
 	maxTXTRecordLength = 255
 	// maxRecordID is the maximum value of record ID (the upper bound for the number
 	// of records with the same type).
-	maxRecordID = 255
+	maxRecordID = 16
```
carpawell, line 63: do we have some ref why this number is good? is it your idea? is it from some original version of NNS?

carpawell (approved):
Did not test.

contracts/nns/contract.go
roman-khimov: No. Yes. No. We definitely need to fix ID overflow problem, but using a lot of records can be a problem performance-wise (likely signifying some design issues). So start small, extend if needed.

contracts/nns/contract.go
roman-khimov: These can add some confusion (similar names) and they're not reused.

contracts/nns/contract.go
```diff
@@ -508,7 +508,7 @@ func SetRecord(name string, typ recordtype.Type, id byte, data string) {
 	if recBytes == nil {
 		panic("invalid record id")
 	}
-	storeRecord(ctx, recordKey, name, typ, id, data)
+	storeRecord(ctx, tokenID, name, typ, id, data)
```
cthulhu-rider, line 520: `getIdRecordKey` redundantly done twice now, not an improvement for me

contracts/nns/contract.go
cthulhu-rider, line 63: shouldnt this be `15` to `limit max number of records to 16`? and i dont think this needs to be separated from 51873d5a20715e6622b373fb2e4a49f634deb417

contracts/nns/contract.go
roman-khimov: This opens up path for 78862e4322fb2b2235a51c4666af94f1fb113a23.

contracts/nns/contract.go
roman-khimov: > shouldnt this be 15

Correct, fixed!

>  i dont think this needs to be separated from 51873d5

Well, that's different from what @AnnaShaleva had originally, so it's either co-authored-by, or this. The last commit also updates binary contracts.

cthulhu-rider: reasonable then

cthulhu-rider (approved):

cthulhu-rider (approved):

PR #420, opened 2024-07-29
NNS update rebased

roman-khimov:
It's a rebase of https://github.com/nspcc-dev/neofs-contract/pull/266 with some additions. It largely intersects with https://github.com/nspcc-dev/neofs-contract/pull/419, but has two additional ones. Not intended to be merged immediately.


roman-khimov:
After #419 merge this completely substitutes #266 with the only two changes left.

carpawell:
Rebase?

roman-khimov:
Doesn't matter much at this stage.

carpawell:
Is it post-release?

roman-khimov:
That's why it's not a part of #419, see above.

carpawell:
I see it (was) based on #419 and has some additional changes but I do not see an explanation as to why not add it to the release. Dont mind though.

roman-khimov:
Because of API changes, https://github.com/nspcc-dev/neofs-contract/pull/266#pullrequestreview-1638131153.

carpawell (commented):
Smth should be done here for sure.

roman-khimov:
The main motivation here is https://github.com/nspcc-dev/neo-go/issues/2795. It's not related to #442/C#, btw, since C# interfaces for these methods are very much different. If we're to solve https://github.com/nspcc-dev/neo-go/issues/2795 this becomes irrelevant, I think.

carpawell (approved):
Does not do things worse.

PR #431, opened 2024-08-29
Go 1.22

roman-khimov:


tests/helpers.go
```diff
@@ -1,16 +1,20 @@
...
+	"math/rand/v2"
 	"testing"
 
 	"github.com/nspcc-dev/neo-go/pkg/neotest"
 	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
 )
 
 func randomBytes(n int) []byte {
+	if n < 8 {
+		panic("too small request")
```
carpawell, line 14: unexpected restriction, comment at least?

carpawell (approved):

tests/helpers.go
roman-khimov: You'll see it immediately if you're to hit it, then you can write a bit more fancy code to deal with it.

Issue #413, opened 2024-06-27
Verification function for container-level thing


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when we can't check something container-level.

## Describe the solution you'd like
Smth. like `VerifySignature(cid interop.Hash256, msg []byte, sigs []Signature, curveHash NamedCurveHash) bool`, we need to:
 * use https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/interop/native/crypto#VerifyWithECDsa internally
 * implement M/N internally precisely following the container placement policy (#412), each REP X should be satisfied by one of the appropriate nodes

This can be used both in contracts and verification scripts (with CID fixed). A set of signatures like this can prove object existence (refs. https://github.com/nspcc-dev/neofs-api/issues/299).

PR #438, opened 2024-10-02
container: add container lists to the contract

### LINKED ISSUE #412: Extend container contract with node lists
I'm always frustrated when we can't easily get a list of container nodes both for GETs of various kinds, but also for signature checks (refs. https://github.com/nspcc-dev/neofs-api/issues/299).

On epoch change IR can recalculate node lists and store them in `container` contract (`netmap` or a separate contracts are also possible, but `container` seems to be a bit more natural). If the list is the same --- nothing should be done. An event should be generated on the list change.

I doubt there are a lot of them if we want to check any container-wide signatures on-chain.

carpawell:
It must be handled by the Alphabet and be updated after every epoch counter increase. Updating is done in two stage (filling a container list and commiting it) to prevent any stack size/memory restrictions. Closes #412.


contracts/container/contract.go
```diff
@@ -491,6 +494,67 @@ func List(owner []byte) [][]byte {
...
+	}
+
+	newNodes := storage.Find(ctx, append([]byte{nextEpochNodesPrefix}, cID...), storage.None)
+	for iterator.Next(newNodes) {
+		newNode := iterator.Value(newNodes).(struct {
+			key []byte
+			val []byte
+		})
+
+		storage.Delete(ctx, newNode.key[:])
```
carpawell, line 541: i spent so much time trying to understand why i was getting an err here but it just refused to work without `[:]`

contracts/container/contract.go
```diff
@@ -491,6 +494,67 @@ func List(owner []byte) [][]byte {
 	return list
 }
 
+// AddNextEpochNodes adds container members for the next epoch. Changes must be
+// committed with [CommitContainerListUpdate]. Without submitting, it is a no-op
```
cthulhu-rider, line 498: i know that u mean, but `no-op` can cause misunderstaning: contract storage is changed, GAS spent, although user sees no side effect through the API. I'd say `AddNextEpochNodes accumulates passed nodes as container members for the next epoch to be committed using [CommitContainerListUpdate].`

tests/container_test.go
```diff
@@ -560,3 +561,48 @@ func requireEstimationsMatch(t *testing.T, expected []estimation, actual []estim
...
+	require.NoError(t, err)
+
+	iter, ok := stack.Pop().Value().(*storage.Iterator)
+	require.True(t, ok)
+
+	fromContract := iteratorToBytesArray(t, iter)
+	require.ElementsMatch(t, containerList, fromContract)
+
+	// clean up container
+	c.Invoke(t, stackitem.Null{}, "commitContainerListUpdate", cID[:])
```
cthulhu-rider, line 587: 1st i'd make one more `addNextEpochNodes` call with `containerList` subset + several completely new elements. And check that old elements are dropped. This is quite non-trivial with add-add-add... semantic sequence

cthulhu-rider (commented):
lux gut overall

@carpawell will there be multiple Add's between Commit's? And might we need deletions?

contracts/container/config.yml
```diff
@@ -1,7 +1,7 @@
 name: "NeoFS Container"
-safemethods: ["alias", "count", "containersOf", "get", "owner", "list", "eACL", "getContainerSize", "listContainerSizes", "iterateContainerSizes", "iterateAllContainerSizes", "version"]
+safemethods: ["alias", "count", "containersOf", "get", "owner", "list", "nodes", "eACL", "getContainerSize", "listContainerSizes", "iterateContainerSizes", "iterateAllContainerSizes", "version"]
 permissions:
-  - methods: ["update", "addKey", "transferX",
+  - methods: ["update", "addKey", "addNextEpochNodes", "commitContainerListUpdate", "transferX",
```
roman-khimov, line 4: This certainly doesn't need to be changed.

contracts/container/contract.go
```diff
@@ -66,6 +67,8 @@ const (
 	containerKeyPrefix   = 'x'
 	ownerKeyPrefix       = 'o'
 	deletedKeyPrefix     = 'd'
+	nodesPrefix          = 'n'
+	nextEpochNodesPrefix = 'c'
```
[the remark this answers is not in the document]
roman-khimov: Yes, prefixes should be distinct. We do have some long ones for historic reasons, but new prefixes must not collide with them.

contracts/container/contract.go
```diff
@@ -491,6 +494,67 @@ func List(owner []byte) [][]byte {
 	return list
 }
 
+// AddNextEpochNodes adds container members for the next epoch. Changes must be
+// committed with [CommitContainerListUpdate]. Without submitting, it is a no-op
+// operation. Results of the call operation can be received via [Nodes]. This
+// method must be called only when a container list is changed, otherwise
+// nothing should be done.
+// Call must be signed by the Alphabet nodes.
+func AddNextEpochNodes(cID []byte, nodes [][]byte) {
```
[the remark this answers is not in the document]
roman-khimov: Can we have just keys there? Everyone knows the network map anyway (can get details based on key). In fact, even a shorter address handle can be used, but this would require a bit more work (so keys are OK).

contracts/container/contract.go
```diff
@@ -491,6 +494,67 @@ func List(owner []byte) [][]byte {
...
+	ctx := storage.GetContext()
+	multiaddr := common.AlphabetAddress()
+	common.CheckAlphabetWitness(multiaddr)
+
+	for _, node := range nodes {
+		pk := node[2 : 2+interop.PublicKeyCompressedLen] // public key in V2 marshaled node info
+		key := []byte{nextEpochNodesPrefix}
+		key = append(key, cID...)
+		key = append(key, pk...)
```
roman-khimov, line 512: 1+32+33 -> looks like an overflow of our key length limit to me (64).

contracts/container/contract.go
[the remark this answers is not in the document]
roman-khimov: Also, please use more specific types. `hash256` is much better than just `[]byte`.

contracts/container/contract.go
```diff
@@ -491,6 +494,67 @@ func List(owner []byte) [][]byte {
...
+			val []byte
+		})
+
+		storage.Delete(ctx, newNode.key[:])
+
+		newKey := append([]byte{nodesPrefix}, newNode.key[1:]...)
+		storage.Put(ctx, newKey, newNode.val)
+	}
+
+	runtime.Notify("ContainerUpdate", cID)
```
roman-khimov, line 547: `NodesUpdate`? We will have mutable containers eventually and here it's not changed.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/mr-tron/base58 v1.2.0
 	github.com/nspcc-dev/neo-go v0.106.3
 	github.com/nspcc-dev/neo-go/pkg/interop v0.0.0-20240729160116-d8e3e57f88f2
+	github.com/nspcc-dev/neofs-sdk-go v1.0.0-rc.12
```
roman-khimov, line 10: I'm not excited about this dependency, can we avoid it?

contracts/container/config.yml
```diff
@@ -28,3 +28,7 @@ events:
     parameters:
       - name: epoch
         type: Integer
+  - name: ContainerUpdate
+    parameters:
+      - name: containerID
+        type: ByteArray
```
roman-khimov, line 34: More specific type, please.

contracts/container/config.yml
```diff
@@ -28,3 +28,7 @@ events:
     parameters:
       - name: epoch
         type: Integer
+  - name: ContainerUpdate
+    parameters:
+      - name: containerID
```
roman-khimov, line 33: Names _are_ usually capitalized for event parameters.

tests/container_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/nspcc-dev/neofs-contract/common"
 	"github.com/nspcc-dev/neofs-contract/contracts/container/containerconst"
 	"github.com/nspcc-dev/neofs-contract/contracts/nns/recordtype"
+	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
```
roman-khimov, line 19: This certainly isn't worth SDK import.

contracts/container/config.yml
carpawell: what do you mean? i am not sure i know the sequence of contract changes. what should be done first, what can be regenerated automatically, what does `permissions` mean?

go.mod
carpawell: sure, only tests use it

contracts/container/config.yml
roman-khimov: Just don't change anything here.

contracts/container/contract.go
roman-khimov: We're risking collisions then and we can't easily extract the public key from the DB key.

contracts/container/contract.go
```diff
@@ -491,6 +494,67 @@ func List(owner []byte) [][]byte {
...
+// method must be called only when a container list is changed, otherwise
+// nothing should be done.
+// Call must be signed by the Alphabet nodes.
+func AddNextEpochNodes(cID []byte, nodes [][]byte) {
+	ctx := storage.GetContext()
+	multiaddr := common.AlphabetAddress()
+	common.CheckAlphabetWitness(multiaddr)
+
+	for _, node := range nodes {
+		pk := node[2 : 2+interop.PublicKeyCompressedLen] // public key in V2 marshaled node info
```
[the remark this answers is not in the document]
carpawell: depends on the final solution about what exact we store in the contract. will change if it is still proto NodeInfo

contracts/container/contract.go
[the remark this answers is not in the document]
carpawell: i tried to fit as much public key in key as possible (cid + pk is already more than contract can have as a key). are there any ideas how to store a list of smth that has 65-byte unique identifier in contract?

contracts/container/contract.go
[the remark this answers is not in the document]
carpawell: > i think we should clarify what exactly these byte arrays are and verify them 1st

> Also, please use more specific types. hash256 is much better than just []byte.

@roman-khimov, if it is `hash256`, can it be ensured by the VM? or still we need to check it manually?

> Can we have just keys there?

we have discussed it a little and i thought that we make it clear that it is useless for current nodes and IR. if you need a container, then you need some ways to communicate with it, not just know its keys. it can be useful for the future things (https://github.com/nspcc-dev/neofs-contract/issues/413), is this PR for contract-side calculations only?

contracts/container/contract.go
```diff
@@ -491,6 +494,67 @@ func List(owner []byte) [][]byte {
...
+	newNodes := storage.Find(ctx, append([]byte{nextEpochNodesPrefix}, cID...), storage.None)
+	for iterator.Next(newNodes) {
+		newNode := iterator.Value(newNodes).(struct {
+			key []byte
+			val []byte
+		})
+
+		storage.Delete(ctx, newNode.key[:])
+
+		newKey := append([]byte{nodesPrefix}, newNode.key[1:]...)
```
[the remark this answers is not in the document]
carpawell: not possible for now (it is a char literally, no length), likely be resolved automatically when we resolve other threads

contracts/container/contract.go
carpawell: i meant without submitting it is not possible to see any changes via public interfaces, it is an alphabet-only method, so gas and any other things are known and does not mean a lot. but don't mind your way

contracts/container/config.yml
carpawell: as usual, i prefer to change things in sync. can you, please, review this repo and create issues about things that are made weird for contracts? i cannot imagine how i should figure it out by myself. as you see, it is "epoch" few lines above

contracts/container/config.yml
carpawell: does not explain the problem. found explanation: https://github.com/nspcc-dev/neo-go/blob/master/docs/compiler.md#permissions
dropped

contracts/container/contract.go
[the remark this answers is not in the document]
roman-khimov: > if it is hash256, can it be ensured by the VM?

No, https://github.com/neo-project/neo/issues/1891.

> if you need a container, then you need some ways to communicate with it

Well, you can get a node by key. And full node info can take quite some space. But let's say it's OK for now.

contracts/container/contract.go
roman-khimov: We need public keys to be accessible for #413, but then the key doesn't mean much and we can use `prefix + cid + counter` or something like that.

contracts/container/contract.go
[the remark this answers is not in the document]
carpawell: oh, this question only about the colision, ok, fixed

contracts/container/config.yml
roman-khimov: OK, I'm wrong here, confused parameter name with event name. See https://github.com/neo-project/proposals/blob/master/nep-17.mediawiki for some reference.

contracts/container/config.yml
roman-khimov: Yes, and the definition we have now sucks a bit ("update" can have a specific contract). But that's not worth fixing now.

contracts/container/contract.go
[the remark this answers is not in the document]
carpawell: > No, https://github.com/neo-project/neo/issues/1891.

fresh thing. ok, type is type, fixed 

> But let's say it's OK for now.

what exacty OK? this PR or storing keys only? i like storing only keys more. just want to be sure we all understand SN and IR will not be using it in the current code, only for new contract features. they will likely use a single network request, a single calculation (placement) and caching, not this PR

contracts/container/contract.go
carpawell: just a question: is a collision (2 bytes less than 33) too scary to complicate things?

> We need public keys to be accessible for #413

pk can be striped on a key suffix and a value. weird but would works and no additional logic with counters and its handling

anyway will fix somehow

contracts/container/contract.go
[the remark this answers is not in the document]
roman-khimov: If we're targeting meta only then full node info isn't needed. But eventually I think this list can be used by other code since no one cares about placement policy, everyone just needs a node list to work with. Full node info can still be provided if this contract is to store keys only, not immediately, but after `netmap` refactorings. So overall keys only are OK.

contracts/container/contract.go
roman-khimov: Probability is low, but the key doesn't make much sense in its current form, it's better to have some meaningful component there.

contracts/container/contract.go
carpawell: @roman-khimov, sorry, wrong line. the problem was in casting kv pair to `struct {[]byte, []byte}` and then i could not change field's first item to `'c'` char with `at instruction 4511 (SETITEM): SETITEM: invalid item type ByteString`. but if i did `struct.key[:][0] = 'c'` it worked no idea why
@AnnaShaleva

carpawell:
> will there be multiple Add's between Commit's?

@cthulhu-rider, yes, if a list is too big we have to make it in more than one call, the stack is limited.

> And might we need deletions?

You mean add-then-delete case? Will it be useful? What cases? New methods should be used by the alphabet nodes once per epoch if some container has changed, not sure I understand if an alphabet node calculates a container, puts it, and then decides to change it in a single epoch.

cthulhu-rider:
> You mean add-then-delete case? Will it be useful? What cases?

best answer is to ask questions huh :] delete question followed after multi-add one. I thought if there is "add" and no "set", where is "delete". Now i see that several "add" are essentially "set" so removals is not needed for now

cthulhu-rider (commented):
lets also update https://github.com/nspcc-dev/neofs-contract/blob/2059417af6e3c98dbcb875cead15cd21db8b1751/contracts/container/doc.go#L72

contracts/container/contract.go
```diff
@@ -491,6 +493,100 @@ func List(owner []byte) [][]byte {
...
+	if len(cID) != interop.Hash256Len {
+		panic(cst.ErrorInvalidContainerID + ": length: " + std.Itoa10(len(cID)))
+	}
+
+	ctx := storage.GetContext()
+	multiaddr := common.AlphabetAddress()
+	common.CheckAlphabetWitness(multiaddr)
+
+	counter := 0
+	counterKey := append([]byte{nextEpochNodesPrefix, 'i'}, cID...)
```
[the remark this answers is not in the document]
carpawell: > isnt this overcomplicated?

maybe. i did not want to have even more prefixes for the same feature, this is my suggestion. can add a dedicated counter prefix if @roman-khimov is also for it

> Collission, although unlikely, is possible

to have problems we need either 31 of 32 bytes should match (with an offset) and the first one should be fixed `i` byte OR a container should have the first 31 bytes of `i`. it is very like "two different containers have the same ID" with almost the same number of problems for NeoFS

contracts/container/contract.go
```diff
@@ -491,6 +493,100 @@ func List(owner []byte) [][]byte {
...
+			val []byte
+		})
+
+		storage.Delete(ctx, newNode.key)
+
+		newKey := append([]byte{nodesPrefix}, newNode.key[1:]...)
+		storage.Put(ctx, newKey, newNode.val)
+	}
+	if newCounter := storage.Get(ctx, newCounterKey); newCounter != nil {
+		storage.Put(ctx, oldCounterKey, storage.Get(ctx, newCounterKey))
```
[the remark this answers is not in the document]
carpawell: > why do we need to keep it?

technically no need, but there is a certain scheme: we have a counter and then _this_ number of keys with counter suffixes. also can imagine requesting current number of container nodes, one Get may be simpler than searching for every entry
tbh, i did not even think about it, just moved all the "new" things to the "old" places. let it be dropped as not needed

contracts/container/contract.go
```diff
@@ -491,6 +493,100 @@ func List(owner []byte) [][]byte {
...
+	} else {
+		storage.Delete(ctx, newCounterKey)
```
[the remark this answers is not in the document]
carpawell: dropped eventually

> and shouldnt we always reset it?

but why not? meaningless counter, yes it may be overwritten with the next container update but before it happenes it is just a garbage

carpawell:
> best answer is to ask questions huh :]

I cannot answer if i do not understand a question (even if it seems like a resolved one now, i still do not understand the original question and its resolution). You asked about deletion, i did not add this so i thought we did not need them. If you have some examples of its usage, i had to ask you to provide them, cause i did not see them. I hope this PR conveys my thoughts and my suggestions, otherwise, i need additional suggestions or change requests.

carpawell:
> lets also update

Hm, did not even know we have it. Updated.

contracts/container/contract.go
roman-khimov: ByteString vs. Buffer, likely.

contracts/container/contract.go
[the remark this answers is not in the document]
roman-khimov: You can avoid storing the counter if you're to take the last element of `nextEpochNodesPrefix + cID` which is easy with `Backwards` flag (https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/interop@v0.0.0-20241007161539-0968c3a81fb0/storage#FindFlags). But to do this you need to have `counter` serialized in BE form.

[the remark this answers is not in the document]
roman-khimov: Or you can avoid using counter with https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/interop@v0.0.0-20241007161539-0968c3a81fb0/contract#CreateStandardAccount. This would solve duplicate problem at the same time.

contracts/container/contract.go
```diff
@@ -491,6 +493,96 @@ func List(owner []byte) [][]byte {
 	return list
 }
 
+// AddNextEpochNodes accumulates passed nodes as container members for the next
+// epoch to be committed using [CommitContainerListUpdate]. Results of the call
+// operation can be received via [Nodes]. This method must be called only when
+// a container list is changed, otherwise nothing should be done.
+// Call must be signed by the Alphabet nodes.
+func AddNextEpochNodes(cID interop.Hash256, publicKeys []interop.PublicKey) {
```
roman-khimov, line 501: How about `REP 1 IN ... REP 2 IN ...`?

roman-khimov: Another thought is managing a _diff_ of lists. But this can be done in future.

contracts/container/contract.go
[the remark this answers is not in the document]
carpawell: @roman-khimov, idea with `Backwards` is the best for me, i like it

> But to do this you need to have counter serialized in BE form.

how can i ensure it? are there any contract side helpers for it?

contracts/container/contract.go
carpawell: > How about REP 1 IN ... REP 2 IN ...?

what do you mean? did you plan to collect signatures by every SELECT? did you plan to have container nodes iterators per SELECTion?

> Another thought is managing a diff of lists

obvious thought but i do not know if it helps. it would complicate IR and contract but i do not see real pros: container change should not be an often operation but even if it happens, a huge number of nodes can be changed at a time with no problems

roman-khimov: How are we going to solve https://github.com/nspcc-dev/neofs-contract/issues/413 without it?

contracts/container/contract.go
```diff
@@ -491,6 +494,143 @@ func List(owner []byte) [][]byte {
 	return list
 }
 
+// AddNextEpochNodes accumulates passed nodes as container members for the next
+// epoch to be committed using [CommitContainerListUpdate]. Nodes must be
+// grouped by selector index from placement policy (SELECT clauses). Results of
+// the call operation can be received via [Nodes]. This method must be called
+// only when a container list is changed, otherwise nothing should be done.
+// Call must be signed by the Alphabet nodes.
+func AddNextEpochNodes(cID interop.Hash256, placementVector int, publicKeys []interop.PublicKey) {
```
roman-khimov, line 503: Should we check the `placementVector`? `add(X, 25, []...)`, for example?

contracts/container/contract.go
```diff
@@ -491,6 +494,143 @@ func List(owner []byte) [][]byte {
...
+	return std.Deserialize(res).(int)
+}
+
+// CommitContainerListUpdate commits container list changes made by
+// [AddNextEpochNodes] calls in advance. Replicas must correspond to
+// ordered placement policy (REP clauses). If no [AddNextEpochNodes]
+// have been made, it clears container list. Makes "ContainerUpdate"
+// notification with container ID after successful list change.
+// Call must be signed by the Alphabet nodes.
+func CommitContainerListUpdate(cID interop.Hash256, replicas []uint8) {
```
roman-khimov, line 559: Better do `[]int`, we don't really separate int types now.

contracts/container/contract.go
```diff
@@ -491,6 +494,143 @@ func List(owner []byte) [][]byte {
...
+		storage.Put(ctx, newKey, newNode.val)
+	}
+
+	rr := storage.Find(ctx, replicasPrefix, storage.KeysOnly)
+	for iterator.Next(rr) {
+		oldReplicasNumber := iterator.Value(rr).([]byte)
+		storage.Delete(ctx, oldReplicasNumber)
+	}
+
+	// nolint:gosimple // this is contract, ranging over nil leads to conversion error
```
roman-khimov, line 597: File an issue for compiler then.

contracts/container/contract.go
```diff
@@ -491,6 +494,143 @@ func List(owner []byte) [][]byte {
...
+		counter++
+
+		storageKey := append(commonPrefix, counterBE(counter)...)
+		storage.Put(ctx, storageKey, publicKey)
+	}
+}
+
+func counterBE(c int) []byte {
+	rawCounter := std.Serialize(c)
+	res := []byte{rawCounter[0], rawCounter[1]} // first is type, second is length
```
roman-khimov, line 536: Why not stripping them completely?

contracts/container/contract.go
```diff
@@ -491,6 +494,143 @@ func List(owner []byte) [][]byte {
...
+	for i := len(rawCounter) - 1; i > 1; i-- {  // LE to BE
+		res = append(res, rawCounter[i])
```
roman-khimov, line 538: We should have fixed result width, otherwise it doesn't work. Suppose 0x01 and 0x0101, we should get `0001` and `0101`, not `01` and `0101`.

contracts/container/contract.go
carpawell: "stripping" == removing? how to deserialize them back then?

contracts/container/contract.go
carpawell: why not considrer it as a part of the interface then? it explains what we expect (even it it does not work)

contracts/container/contract.go
carpawell: > Should we check the placementVector? 

what exact check you mean? size? check that we have 0-24 indexes?

contracts/container/contract.go
```diff
@@ -491,6 +494,143 @@ func List(owner []byte) [][]byte {
...
+	if len(cID) != interop.Hash256Len {
+		panic(cst.ErrorInvalidContainerID + ": length: " + std.Itoa10(len(cID)))
+	}
+
+	ctx := storage.GetContext()
+	multiaddr := common.AlphabetAddress()
+	common.CheckAlphabetWitness(multiaddr)
+
+	commonPrefix := append([]byte{nextEpochNodesPrefix}, cID...)
+	commonPrefix = append(commonPrefix, byte(placementVector))
```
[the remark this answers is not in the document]
carpawell: > Does it work

either tests do not cover it, or it works. can you, please, explain what exact part to recheck? `byte(int)` may work strange? or what slice?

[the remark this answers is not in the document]
roman-khimov: What's the resulting key?

contracts/container/contract.go
roman-khimov: Well, it's not critical, can be prefixed, but a[0]*2^24+a[1]*2^16... I can also show you some magic to avoid `std.Serialize`.

contracts/container/contract.go
roman-khimov: It's the same Integer type in the manifest anyway. Although for NEP-25 there can be some difference, so let's keep it.

contracts/container/contract.go
roman-khimov: 0-24. Maybe it's a bit stretched, but well, it's a contract, it's the last guard before this (broken) data enters the system.

contracts/container/contract.go
carpawell: https://github.com/nspcc-dev/neo-go/issues/3608

carpawell:
@roman-khimov, recheck carefully, please, this is the 4th time i redo the PR, it may have stupid mistakes.

contracts/container/contract.go
roman-khimov: https://github.com/nspcc-dev/neo-go/issues/3610

contracts/container/contract.go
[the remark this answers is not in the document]
roman-khimov: Have you checked it?

contracts/container/contract.go
```diff
@@ -491,6 +494,179 @@ func List(owner []byte) [][]byte {
...
+}
+
+func validatePlacementIndex(ctx storage.Context, cID interop.Hash256, inx uint8) {
+	if inx == 0 {
+		return
+	}
+
+	commonPrefix := append([]byte{nextEpochNodesPrefix}, cID...)
+
+	for i := range int(inx) {
```
roman-khimov, line 576: I'd simplify this to checking `inx-1`, this should be sufficient.

contracts/container/contract.go
```diff
@@ -491,6 +494,179 @@ func List(owner []byte) [][]byte {
...
+
+func counterToBytes(counter int) []byte {
+	var anyCounter any = counter
+	res := anyCounter.([]byte)
+
+	switch len(res) {
+	case 0:
+		res = []byte{0, 0}
+	case 1:
+		res = []byte{res[0], 0}
```
roman-khimov, line 546: `0, res[0]`?

contracts/container/contract.go
```diff
@@ -491,6 +494,179 @@ func List(owner []byte) [][]byte {
...
+	for iterator.Next(rr) {
+		oldReplicasNumber := iterator.Value(rr).([]byte)
+		storage.Delete(ctx, oldReplicasNumber)
+	}
+
+	// nolint:gosimple // https://github.com/nspcc-dev/neo-go/issues/3608
+	if replicas != nil {
+		for i, replica := range replicas {
+			// nolint:staticcheck
+			if replica > 255 {
```
roman-khimov, line 632: Can be a `const`.

contracts/container/contract.go
```diff
@@ -491,6 +494,179 @@ func List(owner []byte) [][]byte {
...
+
+func validatePlacementIndex(ctx storage.Context, cID interop.Hash256, inx uint8) {
+	if inx == 0 {
+		return
+	}
+
+	commonPrefix := append([]byte{nextEpochNodesPrefix}, cID...)
+
+	for i := range int(inx) {
+		iter := storage.Find(ctx, append(commonPrefix, byte(i)), storage.None)
```
roman-khimov, line 577: Does it work with 0?

contracts/container/contract.go
carpawell: dont mind

contracts/container/contract.go
carpawell: should add `return` then. hm, ok

contracts/container/contract.go
carpawell:
```go
runtime.Log("DEBUG: len before: " + std.Itoa10(len(commonPrefix)) + ", len after: " + std.Itoa10(len(append(commonPrefix, byte(0)))))
```

```
"msg": "DEBUG: len before: 33, len after: 34"
```

roman-khimov (approved):

PR #439, opened 2024-10-07
*: enable coverage

### LINKED ISSUE #429: Enable smart contracts coverage
https://github.com/nspcc-dev/neo-go/pull/3462 is almost done, thus coverage is unlocked for NeoFS contracts. We need to:
1. Add make target to collect contracts coverage. Be careful since this repo contains not only contracts, but also simple Go code.
2. Add GithubAction job that will collect and report contracts coverage.
3. Add a coverage badge to README.
4. Profit! Space for contracts coverage improvement is created.

AnnaShaleva:
WIP, waiting for the next NeoGo release to upgrade, but I have this code locally, so anyway it will be helpful.

Close #429.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -34,6 +34,33 @@ jobs:
     - name: Test
       run: go test -v ./...
 
+  cover:
+    name: Coverage
+    runs-on: ubuntu-latest
+
+    env:
+      CGO_ENABLED: 0
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.23'
+          cache: true
+
+      - name: Write coverage profile
+        run: go test -v ./... -coverprofile=./coverage.txt -coverpkg=./contracts/...
+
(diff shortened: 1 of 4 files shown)


.github/workflows/go.yml
```diff
@@ -34,6 +34,34 @@ jobs:
     - name: Test
       run: go test -v ./...
 
+  cover:
+    name: Coverage
+    runs-on: ubuntu-22.04
```
roman-khimov, line 39: latest?

.github/workflows/go.yml
```diff
@@ -34,6 +34,34 @@ jobs:
...
+
+    env:
+      CGO_ENABLED: 0
+      GOEXPERIMENT: nocoverageredesign
```
roman-khimov, line 43: Likely can be avoided here.

AnnaShaleva: Can't.

AnnaShaleva: Well, it can't on my machine. But this action uses Go 1.23, so may be we should try.

roman-khimov (approved):

PR #442, opened 2024-10-31
NNS changes from C#

roman-khimov:
This updates and replaces #267. Most of the changes are already merged via other PRs, but:
 * #420 changes are duplicated in #267 and so are also present here
 * domain registration price/length dependency is omitted from this set because it doesn't fit our model at all

Otherwise they're OK to stay here for now (binaries/bindings not updated till we make decisions).


contracts/nns/contract.go
```diff
@@ -629,7 +629,8 @@ func Resolve(name string, typ recordtype.Type) []string {
 	}
 
 	ctx := storage.GetReadOnlyContext()
-	return resolve(ctx, nil, name, typ, 2)
+	res := []string{}
```
cthulhu-rider, line 652: the compiler understands inline nil but not empty slice?

contracts/nns/contract.go
```diff
@@ -461,16 +463,30 @@ func saveDomain(ctx storage.Context, name, email string, refresh, retry, expire,
 	putSoaRecord(ctx, name, email, refresh, retry, expire, ttl)
 }
 
-// Renew increases domain expiration date.
-func Renew(name string) int64 {
+// RenewDefault increases domain expiration date for 1 year.
```
cthulhu-rider, line 466: and returns what?

contracts/nns/contract.go
```diff
@@ -461,16 +463,30 @@ func saveDomain(ctx storage.Context, name, email string, refresh, retry, expire,
...
+func RenewDefault(name string) int64 {
+	return Renew(name, 1)
+}
+
+// Renew increases domain expiration date up to the specified amount of years.
```
cthulhu-rider, line 471: same

contracts/nns/contract.go
```diff
@@ -461,16 +463,30 @@ func saveDomain(ctx storage.Context, name, email string, refresh, retry, expire,
...
+func Renew(name string, years int) int64 {
+	if years < 1 || years > 10 {
+		panic("invalid renewal period value")
```
cthulhu-rider, line 474: limit not mentioned in docs

cthulhu-rider (commented):
good overall

contracts/nns/contract.go
carpawell: hm, yes, why not just `[]string{}` as an arg?

carpawell (dismissed):

roman-khimov:
Let's merge #420 then, some changes are present there and if we don't have any questions there it'd simplify things a bit.

contracts/nns/contract.go
roman-khimov: Fixed in #420.

carpawell:
@roman-khimov, done, can be rebased, also linter asks for attention.

carpawell (approved):

cthulhu-rider (approved):

PR #448, opened 2024-11-27
container: add object meta notification support

### LINKED ISSUE #414: Per-container event feed
I'm always frustrated when I don't know what's up with the container contents. And we want to filter events per-container. 

The solution is to deploy a contract for containers that use meta (https://github.com/nspcc-dev/neofs-api/issues/300, container contract would need to know it's enabled). This contract needs to accept event data from SNs and push them into notifications. It does not need to process/store these events. This gives a natural event feed for container.

carpawell:
Containers now can be created with new `put` method with one more bool parameter for meta-on-chain feature (TX must still be signed by the Alphabet, and it also must check that container was created with this feature, see https://github.com/nspcc-dev/neofs-api/pull/309). If enabled, new `PutObject` method allows checking object meta information's signature and producing corresponding notification. Closes #414.


contracts/container/contract.go
```diff
@@ -238,6 +239,42 @@ func Update(script []byte, manifest []byte, data any) {
...
+	if storage.Get(storage.GetContext(), append([]byte{containersWithMetaPrefix}, cID...)) == nil {
+		panic("container does not support meta-on-chain")
+	}
+
+	if !VerifyPlacementSignatures(cID, metaInformation, sigs) {
+		panic("signature verification failed")
+	}
+
+	oid := metaMap["oid"].(interop.Hash256)
+	runtime.Notify("ObjectPut", cID, oid)
```
carpawell, line 263: @roman-khimov, see if it suits your expectations about notifications. on one hand, it is ok to know only cid/oid, on the other hand, why we have this info if it is not pushed to the notification? but also, we have a limitation on notification length, so no 1000 object deletion at a time and notification about it. also, we may have more fields in the future, should we change notification interface then?

contracts/container/contract.go
```diff
@@ -238,6 +239,42 @@ func Update(script []byte, manifest []byte, data any) {
...
+	cID := metaMap["cid"].(interop.Hash256)
+	if len(cID) != interop.Hash256Len {
+		panic("incorrect container ID")
+	}
+
+	if storage.Get(storage.GetContext(), append([]byte{containersWithMetaPrefix}, cID...)) == nil {
+		panic("container does not support meta-on-chain")
+	}
+
+	if !VerifyPlacementSignatures(cID, metaInformation, sigs) {
```
roman-khimov, line 258: Current execution bottleneck for sure. But can't be solved with the interface we have for now.

contracts/container/contract.go
```diff
@@ -238,6 +239,42 @@ func Update(script []byte, manifest []byte, data any) {
...
+		panic("signature verification failed")
+	}
+
+	oid := metaMap["oid"].(interop.Hash256)
```
roman-khimov, line 262: Should be verified too.

contracts/container/contract.go
roman-khimov: You have this call per event (object) anyway. And we need max data possible, so maybe pushing the whole map is appropriate here. 1K size is per even also, so it won't be a problem.

contracts/container/contract.go
```diff
@@ -238,6 +239,42 @@ func Update(script []byte, manifest []byte, data any) {
...
+	runtime.Notify("ObjectPut", cID, oid)
+}
+
+// PutMeta is the same as [Put] (and exposed as put from the contract via
+// overload), but allows container's meta-information be handled and notified
+// using the chain.
+func PutMeta(container []byte, signature interop.Signature, publicKey interop.PublicKey, token []byte, metaOnChain bool) {
```
roman-khimov, line 269: Related to https://github.com/nspcc-dev/neofs-contract/issues/449. You can drop `metaOnChain` parameter, if this method is called --- it's meta-enabled.

contracts/container/contract.go
```diff
@@ -61,15 +61,16 @@ const (
...
-	containerKeyPrefix   = 'x'
-	ownerKeyPrefix       = 'o'
-	deletedKeyPrefix     = 'd'
-	nodesPrefix          = 'n'
-	replicasNumberPrefix = 'r'
-	nextEpochNodesPrefix = 'u'
-	estimatePostfixSize  = 10
+	singleEstimatePrefix     = "est"
+	estimateKeyPrefix        = "cnr"
+	containersWithMetaPrefix = 'm'
```
cthulhu-rider, line 66: just a remark

i very like (no) how gofmt makes a 10-line change when one is actually added/changed. Takin bout why linebreaks are good sometimes, and const/var blocks aint

contracts/container/contract.go
carpawell: > 1K size is per even also, so it won't be a problem.

but we can delete 1k objects with a single tombstone object, this is 1k hashes according to https://github.com/nspcc-dev/neofs-node/blob/7b592bdb1636bb67f55100877b3bf72fcc43ff4a/pkg/core/object/replicate.go#L35

contracts/container/contract.go
carpawell: i did it overloaded so it can be used as `Put(..., cnr.METAINFO_CONSISTENCY == "strict"`)

we can make #449 before this PR (IMO, a preferable way)

contracts/container/contract.go
carpawell: you mean length? agree it should be verified but it is not used for the contract so i checked only the essentials. if we want to check oid, i would say we need to check all the map somehow. but the issue says: "It does not need to process/store these events"

contracts/container/contract.go
carpawell: > Current execution bottleneck for sure

what (potential) improvement you meat?

contracts/container/contract.go
roman-khimov: > process/store

I mean creating any indexes/DB. But some sanity checks are better be performed.

contracts/container/contract.go
roman-khimov: Signing a transaction.

contracts/container/contract.go
roman-khimov: We need more strict limits there then. We no longer _need_ this for split objects, so I'd even think of 1:1 mapping (single tombstone --- single object).

contracts/container/contract.go
```diff
@@ -238,6 +239,42 @@ func Update(script []byte, manifest []byte, data any) {
 	runtime.Log("container contract updated")
 }
 
+// PutObject registers successful object PUT operation and notifies about
+// it. metaInformation must be signed by container nodes according to
+// container's placement, see [VerifyPlacementSignatures]. metaInformation
+// must contain information about an object placed to a container that was
+// created with meta-on-chain feature enabled.
+func PutObject(metaInformation []byte, sigs [][]interop.Signature) {
```
cthulhu-rider, line 247: not the best name for the semantics imo, suggest Notify/Submit/Handle + ObjectPut

tests/container_test.go
```diff
@@ -804,3 +804,44 @@ func stackWithBool(t *testing.T, stack *vm.Stack, v bool) {
...
+		acc := c.NewAccount(t)
+		cnt := dummyContainer(acc)
+		balanceMint(t, cBal, acc, containerFee*1, []byte{})
+
+		metaInfo, err := stackitem.Serialize(stackitem.NewMapWithValue(
+			[]stackitem.MapElement{{Key: stackitem.Make("cid"), Value: stackitem.Make(cnt.id[:])}}))
+		require.NoError(t, err)
+
+		c.Invoke(t, stackitem.Null{}, "put", cnt.value, cnt.sig, cnt.pub, cnt.token, false)
+		c.InvokeFail(t, "container does not support meta-on-chain", "putObject", metaInfo, nil)
```
cthulhu-rider, line 821: lets cover all cases:
* invalid meta info (not a map, invalid cid/oid)
* invalid signatures

tests/container_test.go
```diff
@@ -804,3 +804,44 @@ func stackWithBool(t *testing.T, stack *vm.Stack, v bool) {
...
+
+		metaInfo, err := stackitem.Serialize(stackitem.NewMapWithValue(
+			[]stackitem.MapElement{
+				{Key: stackitem.Make("cid"), Value: stackitem.Make(cnt.id[:])},
+				{Key: stackitem.Make("oid"), Value: stackitem.Make(cnt.id[:])},
+			}))
+		require.NoError(t, err)
+
+		c.Invoke(t, stackitem.Null{}, "put", cnt.value, cnt.sig, cnt.pub, cnt.token, true)
+		c.Invoke(t, stackitem.Null{}, "putObject", metaInfo, nil)
```
cthulhu-rider, line 845: why is it OK w/o signatures?

cthulhu-rider, line 845: if we have tools to test whether notification event is expectedly thrown, lets do it. If no, suggest to create an issue for this. Suppose other methods/contracts need this too

contracts/container/contract.go
cthulhu-rider, line 269: we put container and just configure meta support, with this `PutMeta` name can confuse. I can suggest `PutWithTunedMeta`, but not very against ur proposal

contracts/container/contract.go
```diff
@@ -238,6 +239,42 @@ func Update(script []byte, manifest []byte, data any) {
 	runtime.Log("container contract updated")
 }
 
+// PutObject registers successful object PUT operation and notifies about
```
cthulhu-rider, line 242: what meaning do u put into `registers` here? In proposed implementation it verifies and notifies only

contracts/container/contract.go
```diff
@@ -238,6 +239,42 @@ func Update(script []byte, manifest []byte, data any) {
...
+// it. metaInformation must be signed by container nodes according to
+// container's placement, see [VerifyPlacementSignatures]. metaInformation
+// must contain information about an object placed to a container that was
+// created with meta-on-chain feature enabled.
```
cthulhu-rider, line 246:
```suggestion
// created using [Put] ([PutMeta]) with enabled meta-on-chain option.
```

contracts/container/contract.go
carpawell: https://github.com/nspcc-dev/neofs-api/issues/311

contracts/container/contract.go
carpawell: > what meaning do u put

dont remember, maybe i meant how it really will be used: some meta service will wait for this notification. this is like the proof that an object exists. until this method is called it is not guaranteed

contracts/container/contract.go
carpawell: i have adopted @roman-khimov's [approach](https://github.com/nspcc-dev/neofs-contract/blob/f6398fe878a1a4285f522ca2d433781812c6bba4/contracts/netmap/contract.go#L497) with method overloading. no one will call it `putMeta`, it is the same `put` but with one more arg. i really liked it

tests/container_test.go
carpawell: we use neo-go's test framework. i tried my best, @roman-khimov, check if there is more optimal way to test events

carpawell: for test purposes, container was added without any nodes (again, IMO signature verification should not be tested in this PR), so no nodes can add its signature but also no nodes are quired to do it, every iterator over node's put keys does nothing

tests/container_test.go
carpawell: > invalid meta info (not a map, invalid cid/oid)

added

> invalid signatures

can we skip it? it literally just calls `VerifyPlacementSignatures` from the same contract and it is already tested in https://github.com/nspcc-dev/neofs-contract/pull/440 and i have nothing to add to it

contracts/container/contract.go
```diff
@@ -238,6 +239,42 @@ func Update(script []byte, manifest []byte, data any) {
 	runtime.Log("container contract updated")
 }
 
+// PutObject registers successful object PUT operation and notifies about
+// it. metaInformation must be signed by container nodes according to
+// container's placement, see [VerifyPlacementSignatures]. metaInformation
+// must contain information about an object placed to a container that was
```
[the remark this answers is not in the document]
carpawell: > what form

i think this will somehow appear somewhere. at least it should appear before node's 0.44.0 release and contract's 0.21. however, currently node also uses it as is and the format is not fixed or described

contracts/container/contract.go
cthulhu-rider: `put` but with one more arg - `putWithMeta` to me

cthulhu-rider (approved):

roman-khimov (approved):
OK for now.

PR #451, opened 2024-12-11
Fix/vub for meta on chain and method overloads

carpawell:


contracts/container/contract.go
```diff
@@ -277,7 +277,7 @@ func SubmitObjectPut(metaInformation []byte, sigs [][]interop.Signature) {
 		}
 	}
 	vub := getFromMap(metaMap, "validuntil").(int)
-	if vub <= ledger.CurrentIndex() {
+	if vub <= contract.Call(storage.Get(ctx, netmapContractKey).(interop.Hash160), "epoch", contract.ReadOnly).(int) {
```
roman-khimov, line 280: I'd still expect blocks here. Epochs are too coarse.

roman-khimov: I know you have problems calculating it on different nodes, but this can be solved in different ways.

roman-khimov: And epochs are not much different here.

carpawell:
@roman-khimov, I do not like the second commit, but this is how we can fix it. Now, I know that overloads exist, and I want them all to be overloaded, not like they are partly now. But to be fully honest, `put` and `putNamed` are the same method;`"", ""` makes them the same. However, with this PR, it is more correct, but I do not like it still.

contracts/container/contract.go
carpawell: > I'd still expect blocks here. Epochs are too coarse.

Well, maybe. Still, we need to sync them somehow.

> I know you have problems calculating it on different nodes, but this can be solved in different ways.

Like the last block in the current epoch + const? I don't mind but I have nothing to inspire me here. Epoches are just a little bit more closer to our storage.

> And epochs are not much different here.

In the node it is currently creation epoch + const.

roman-khimov: But if it's in blocks then that's the value to use, isn't it? One node sets it, others are either OK or not with it (usually they're OK).

roman-khimov:
Regarding overloads -- it's ok for me for now.

roman-khimov:
In general, overloaded `put` would be better, but we already use `putNamed` a lot, it can't be changed easily.

contracts/container/contract.go
roman-khimov: My suggestion is to drop this patch.

roman-khimov:
Although you can have proper set of puts and putNamed at the same time.

contracts/container/contract.go
carpawell: > My suggestion is to drop this patch.

i don't mind but what block to use in the nodes then? i can only think of https://github.com/nspcc-dev/neofs-contract/blob/f5a31fd10a31988b019feaa6a04aeaf27f71afbf/contracts/netmap/contract.go#L454 + const and hope it is not gonna tick while signatures are exchanged

roman-khimov: You can use current + something, it is exchanged between nodes, so it's not a problem for other nodes to use the value.

contracts/container/contract.go
```diff
@@ -296,16 +296,16 @@ func getFromMap(m map[string]any, key string) any {
 	return m[key]
 }
 
-// PutMeta is the same as [Put] (and exposed as put from the contract via
+// PutMeta is the same as [PutNamed] (and exposed as put from the contract via
```
cthulhu-rider, line 299: does this make impossible to `put` with `metaOnChain`?

carpawell: we should either duplicate code, or be ready to do only one thing in PUTNAMED/PUT pair. _but_ PUT still can be done with PUTNAMED with "", "" name and zone

cthulhu-rider: didnt get the answer. With proposed changes, can i `put(container, signature, publicKey, token, true)`?

carpawell: > does this make impossible to put with metaOnChain?

yes, it makes. we either have dozen of methods to overload every case or use the fact that "", "" args make PutNamed behave as Put

> With proposed changes, can i put(container, signature, publicKey, token, true)?

no. but before you were impossible to create a named container with meta-on-chain. now you can `put(container, signature, publicKey, token, "", "", true)` and it is the same as your request but you need to understand it and put such empty args. i was writing code near this method calls when i realized it. now i want that code to be always "put" and nothing more. it simplifies things

carpawell:
First commit dropped BTW.

contracts/container/contract.go
```diff
@@ -320,8 +322,15 @@ func Put(container []byte, signature interop.Signature, publicKey interop.Public
...
 
+// PutNamedOverloaded is the same as [Put] (and exposed as put from the contract via
+// overload), but allows named container creation via NNS contract.
+func PutNamedOverloaded(container []byte, signature interop.Signature, publicKey interop.PublicKey, token []byte, name, zone string) {
+	PutNamed(container, signature, publicKey, token, name, zone)
+}
+
 // PutNamed is similar to put but also sets a TXT record in nns contract.
 // Note that zone must exist.
+// DEPRECATED: use [PutNamedOverloaded].
```
cthulhu-rider, line 333: iiuc from https://github.com/nspcc-dev/neofs-contract/pull/451#issuecomment-2538854815, `put` is for everything. So, i'd deprecate this favor of `Put`, not `PutNamedOverloaded`

it's also interesting do contract manifests support `deprecated` markers? @roman-khimov

contracts/container/contract.go
cthulhu-rider: my proposal is already implemented https://github.com/nspcc-dev/neofs-contract/pull/451#issuecomment-2538854815

contracts/container/contract.go
carpawell: > So, i'd deprecate this favor of Put

well, maybe. depends on how to read the code (methods that are written here will never be called on client side, it will be just "put"). hm, since we need to write here something, let it be ": call [Put] with the same args instead"

roman-khimov (approved):

PR #457, opened 2024-12-23
Feat/support meta with big objects

carpawell:


roman-khimov (commented):
How about previous ID?

carpawell:
@roman-khimov, PR became a better version of itself.

roman-khimov (approved):

PR #461, opened 2025-01-22
netmap: add a temporary usev2 flag

roman-khimov:
Use it for any new installations, but hold back old ones till they're migrated.


cthulhu-rider (approved):
code is ok

pls describe a motivation in the commit, why it's needed, how it's gonna be used and, if it is temporary, when will go

carpawell:
Agree, the purpose is not clear.

roman-khimov:
The purpose is https://github.com/nspcc-dev/neofs-node/pull/3088. And also to have new map only for new installations.

roman-khimov:
It can technically work without zero pushed into the DB on update, but one for new installations is important enough to me, I don't want to have legacy being used for new installations.

PR #462, opened 2025-01-24
netmap: expose Candidate structure to RPC bindings

roman-khimov:
Unfortunately, we don't have any more elegant way to do this now.


cthulhu-rider (approved):
changes ok, didnt get why exactly this is need, but sure it is

roman-khimov:
Can't use `Candidate.FromStackItem` otherwise, it's just missing from the RPC code.

PR #464, opened 2025-02-06
netmap: add IsStorageNode() for the container contract

roman-khimov:
1. Container contract shouldn't parse BLOBs.
2. It shouldn't also iterate over the whole map, this check is much easier to perform for v2 nodes.
3. The check using v1 only is obviously broken for a v2 only map, so this fixes payment tests.


contracts/netmap/contract.go
```diff
@@ -546,6 +546,51 @@ func ListCandidates() iterator.Iterator {
 		storage.ValuesOnly|storage.DeserializeValues)
 }
 
+// IsStorageNode allows to check for the given key presence in the current
+// network map.
+func IsStorageNode(key interop.PublicKey) bool {
+	return IsStorageNodeInEpoch(key, Epoch())
```
carpawell, line 552: calling this method does two `Epoch()` calls, do not think it is bad but still

carpawell (approved):

contracts/netmap/contract.go
roman-khimov: It's cheap.

PR #478, opened 2025-04-09
Adapt container API to N3 account auth scheme

cthulhu-rider:


contracts/container/config.yml
```diff
@@ -9,7 +9,7 @@ events:
       - name: containerID
         type: Hash256
       - name: publicKey
-        type: PublicKey
+        type: Hash160
```
roman-khimov, line 12: We need some new events, changing signature of old ones is questionable.

contracts/container/config.yml
```diff
@@ -9,7 +9,7 @@ events:
       - name: containerID
         type: Hash256
       - name: publicKey
```
roman-khimov, line 11: Then this is no longer a `publicKey`, rather an `owner`.

contracts/container/contract.go
```diff
@@ -444,7 +444,9 @@ func PutNamed(container []byte, signature interop.Signature,
 	}
 
 	runtime.Log("added new container")
-	runtime.Notify("PutSuccess", containerID, publicKey)
+	h := crypto.Sha256(publicKey)
+	acc := crypto.Ripemd160(h[:])
```
roman-khimov, line 448: Better just emit its owner. `publicKey` can be null eventually, I think we can avoid using it.

cthulhu-rider: owner diffs from signer when session token is provided. We wanna throw owner in this case too?

contracts/container/config.yml
cthulhu-rider: kept the old event

contracts/container/contract.go
cthulhu-rider: done with container owner

contracts/container/config.yml
```diff
@@ -9,7 +9,7 @@ events:
...
-        type: PublicKey
+        type: ByteArray
```
roman-khimov, line 12: Can this change be avoided?

cthulhu-rider: i tried passing N3 verification and invocation scripts as public key and signature parameters correspondingly. Verification script was 47 byte len. Faced following error:
```
System.Runtime.Notify failed: notification PutSuccess is invalid: parameter 1 type mismatch: PublicKey (manifest) vs ByteString (notification)
```
so i decided that this extension is necessary. Is there a better way to get around this with min changes to the contract?

contracts/container/contract.go
```diff
@@ -447,6 +450,93 @@ func PutNamed(container []byte, signature interop.Signature,
...
+	if name != "" {
+		fee += contract.Call(netmapContract, "config", contract.ReadOnly, cst.AliasFeeKey).(int)
+	}
+
+	if ownerBalance < fee*len(alphabet) {
+		panic("insufficient balance to create container")
+	}
+
+	transferDetails := common.ContainerFeeTransferDetails(id)
+	for i := range alphabet {
```
roman-khimov, line 526: Can be avoided if `fee == 0`.

contracts/container/contract.go
```diff
@@ -501,6 +592,47 @@ func Delete(containerID []byte, signature interop.Signature, token []byte) {
...
+	ctx := storage.GetContext()
+	cnrItemKey := append([]byte{containerKeyPrefix}, id...)
+	cnrItem := storage.Get(ctx, cnrItemKey)
+	if cnrItem == nil {
+		return
+	}
+
+	owner := ownerFromBinaryContainer(std.Deserialize(cnrItem.([]byte)).(Container).Value)
+
+	storage.Delete(ctx, cnrItemKey)
```
roman-khimov, line 616: Why not reusing `removeContainer` here?

contracts/container/contract.go
cthulhu-rider: done for the new method

contracts/container/contract.go
cthulhu-rider: taking into account `Delete` is going to be removed, `removeContainer` is used once. I also noticed that when it's inlined, contract's NEF and manifest change (understandable but surprising still). And to me it's easier to see which storage items are affected w/o jumps. So i propose to keep as is, but dont mind to change

PR #486, opened 2025-04-17
netmap: Persist epoch tick timings

cthulhu-rider:


contracts/netmap/contract.go
```diff
@@ -421,7 +428,8 @@ func NewEpoch(epochNum int) {
 
 	// todo: check if provided epoch number is bigger than current
 	storage.Put(ctx, snapshotEpoch, epochNum)
-	storage.Put(ctx, snapshotBlockKey, ledger.CurrentIndex())
+	curIndex := ledger.CurrentIndex()
+	storage.Put(ctx, snapshotBlockKey, curIndex)
```
roman-khimov, line 432: Do we need this key then? It can be derived from the new index.

contracts/netmap/contract.go
```diff
@@ -790,6 +802,18 @@ func UnusedCandidate() Candidate {
...
 }
 
+// GetEpochBlock returns block index when given epoch came. Returns -1 if the
+// epoch is missing.
+//
+// Use [LastEpochBlock] if you are interested in the current epoch.
+func GetEpochBlock(epoch int) int {
+	val := storage.Get(storage.GetReadOnlyContext(), append([]byte{epochIndexKey}, fourBytesBE(epoch)...))
+	if val == nil {
+		return -1
```
roman-khimov, line 812: Why not zero? It's never a valid block for an epoch as well. We use unsigned types for block numbers a lot and this can be confusing.

contracts/netmap/contract.go
```diff
@@ -442,6 +450,8 @@ func NewEpoch(epochNum int) {
 	// make clean up routines in other contracts
 	cleanup(ctx, epochNum)
 
+	storage.Put(ctx, append([]byte{epochIndexKey}, fourBytesBE(epochNum)...), curIndex)
```
roman-khimov, line 453: How about storing the time in the same index? We don't have use cases for it right now, but it's not hard to add and it can be useful in future if we're to try doing something time-dependent.

contracts/netmap/migration_test.go
```diff
@@ -184,7 +181,11 @@ func testMigrationFromDump(t *testing.T, d *dump.Reader) {
 
 	c.CheckUpdateSuccess(t, updPrm...)
 
-	checkNewEpochSubscribers(t, c, balanceHash, containerHash)
+	if prevVersion < 19_000 {
```
roman-khimov, line 184: A part of the problem is that these new CVSs are _too new_. Migrations we have currently in the code are somewhat trivial (maybe except for the balance contract), nothing compared to the notary migration.

I though about picking some older (post-notary, but not the latest) state, but never managed to do that.

contracts/netmap/contract.go
cthulhu-rider: technically we dont, it's accessible via 2 `storage.Get` ops. Worth dropping?

contracts/netmap/contract.go
cthulhu-rider: zero seems better

how about declaring `uint32` return?

contracts/netmap/migration_test.go
cthulhu-rider: > I though about picking some older (post-notary, but not the latest) state

i thought about this too. Can we conveniently determine the moments at which past updates were made? It seems that such a chronology can be useful, similar to epochs. But apparently possible to restore it only manually now

contracts/netmap/contract.go
roman-khimov: `uint32` won't help much until we have NEP-25 properly implemented, so don't bother for now.

contracts/netmap/contract.go
roman-khimov: It's not often used and for an RPC invocation the difference in access time is nothing. But it's one key less in the DB, always good.

contracts/netmap/contract.go
roman-khimov: Yes, structure is easier to handle and can be extended in future if need be.

contracts/netmap/contract.go
cthulhu-rider: done

contracts/netmap/contract.go
cthulhu-rider: switched to zero

contracts/netmap/contract.go
cthulhu-rider: done

contracts/netmap/migration_test.go
cthulhu-rider: suggest resolving in a separate issue

roman-khimov: We can just forget about it for a moment, all of the current (pre-0.22.0) migrations are basically done and new one is tested here, so not a problem.

contracts/netmap/contract.go
```diff
@@ -790,6 +810,18 @@ func UnusedCandidate() Candidate {
 	return Candidate{}
 }
 
+// GetEpochBlock returns block index when given epoch came. Returns 0 if the
+// epoch is missing.
+//
+// Use [LastEpochBlock] if you are interested in the current epoch.
+func GetEpochBlock(epoch int) int {
```
roman-khimov, line 817: GetEpochTime as well?

cthulhu-rider: will add. Can `LastEpochTime` be also useful?

roman-khimov: I think so. In fact, I'd really like to switch epochs from blocks to time. It can be done and we'd need this time data then.

roman-khimov: At the moment let's just expose everything we have from the contract and then we'll figure out how to use or abuse it.

cthulhu-rider: added them both

contracts/netmap/contract.go
```diff
@@ -178,6 +185,13 @@ func _deploy(data any, isUpdate bool) {
 			setConfig(ctx, "UseNodeV2", []byte{0})
 		}
 
+		if version < 22_000 {
+			curEpoch := storage.Get(ctx, snapshotEpoch).(int)
+			curEpochHeight := storage.Get(ctx, snapshotBlockKey).(int)
+			storage.Put(ctx, append([]byte{epochIndexKey}, fourBytesBE(curEpoch)...), std.Serialize(epochItem{height: curEpochHeight}))
```
roman-khimov, line 191: Time info isn't present (can be retrieved from block header), but I doubt it's critical, the next epoch will have it anyway.

roman-khimov (approved):

Issue #459, opened 2025-01-11
Support NEP-22


roman-khimov:
## Is your feature request related to a problem? Please describe.
I'm always frustrated when we're not using standard methods of doing something when they exist.

## Describe the solution you'd like
NEP-22 was introduced recently and we should check if we're compliant (likely we are in fact) and declare this in manifest.

## Additional context
https://github.com/neo-project/proposals/blob/master/nep-22.mediawiki
https://github.com/nspcc-dev/neo-go/issues/3784

cthulhu-rider:
is it important for the parameter name to match the standard?

[NEP](https://github.com/neo-project/proposals/blob/master/nep-22.mediawiki#update)
```json
{
  "name": "update",
  "safe": false,
  "parameters": [
    {
      "name": "nefFile",
      "type": "ByteArray"
    },
    {
      "name": "manifest",
      "type": "ByteArray"
    },
    {
      "name": "data",
      "type": "Any"
    }
  ],
  "returntype": "Void"
}
```
vs, in Alphabet [manifest](https://github.com/nspcc-dev/neofs-contract/blob/v0.21.0/contracts/alphabet/manifest.json) for example, 
```json
{
  "name": "update",
  "offset": 1247,
  "parameters": [
    {
      "name": "script",
      "type": "ByteArray"
    },
    {
      "name": "manifest",
      "type": "ByteArray"
    },
    {
      "name": "data",
      "type": "Any"
    }
  ],
  "returntype": "Void",
  "safe": false
}
```

PR #490, opened 2025-04-23
Follow NEP-22

cthulhu-rider:
while doing https://github.com/nspcc-dev/neofs-contract/issues/459#issuecomment-2824199031, i noticed that 2nd `_deploy` prm has `isUpdate` name in out manifests while https://github.com/neo-project/proposals/blob/master/nep-29.mediawiki declares it as `update`. Tried to rename it in code and re-compile, but manifests did not change. Apparently it's built into the compiler and cannot be changed https://github.com/nspcc-dev/neo-go/blob/3b54213c08ab714cb9be479fd0a7312463e0e3c0/pkg/compiler/debug.go#L194


contracts/alphabet/contract.go
```diff
@@ -68,7 +68,7 @@ func _deploy(data any, isUpdate bool) {
 
 // Update method updates contract source code and manifest. It can be invoked
 // only by committee.
-func Update(nefFile []byte, manifest []byte, data any) {
+func Update(nefFile, manifest []byte, data any) {
```
roman-khimov, line 71: I don't like this style since it breaks simple "name type" sequence, you have to walk past the comma to figure out what's going on. I can live with it though.

roman-khimov (commented):
You need to add "NEP-22" to supported standards as well.

carpawell (approved):
Just a curiosity: do you think it is a good idea to ref first commits in the further ones? I see how you do it sometimes, IMO it is not safe and sooner or later we will merge a link to a nonexistent commit.

roman-khimov (approved):

PR #494, opened 2025-06-02
proxy,container: do not collect notary signature for metadata

### LINKED ISSUE #491: Pay for metadata on chain by proxy contract without Alphabet signature
I'm always frustrated when we need to sponsor our system things but an Alphabet signature is required for this: https://github.com/nspcc-dev/neofs-contract/blob/d490c0d5624db991da4166cf7fb347056749b03b/contracts/alphabet/contract.go#L214

Make an exception for hot cases. [SubmitObjectPut](https://github.com/nspcc-dev/neofs-contract/blob/d490c0d5624db991da4166cf7fb347056749b03b/contracts/container/contract.go#L197) may be such. It is possible to allow paying for invocations that are signed by our contracts, and contracts may have verification logic on their side.

Notary flow. Additional latency, additional load on IR nodes.

## Additional context
https://github.com/nspcc-dev/neofs-node/pull/3322

carpawell:
Make an exception for `SubmitObjectPut` in Proxy's `Verify` method. Closes #491.


contracts/proxy/contract.go
```diff
@@ -42,8 +53,69 @@ func Update(nefFile, manifest []byte, data any) {
 
 // Verify checks whether carrier transaction contains either (2/3N + 1) or
 // (N/2 + 1) valid multi-signature of the NeoFS Alphabet.
+// Container contract's `SubmitObjectPut` is an exception, Alphabet signature
+// is not required, contract's `VerifyPlacementSignatures` is called instead.
 func Verify() bool {
-	return common.ContainsAlphabetWitness()
+	cnrH := storage.Get(storage.GetReadOnlyContext(), containerContractKey).(interop.Hash160)
```
AnnaShaleva, line 59: Can't we use NNS to resolve container contract hash?

carpawell: in fact, it would simplify many-many things. i need an approve that it is not a bad idea compared to `storage.Get`

carpawell: @roman-khimov

contracts/proxy/contract.go
```diff
@@ -42,8 +53,69 @@ func Update(nefFile, manifest []byte, data any) {
...
+		if scriptArgsOnly[i] <= 5 { // PUSHINT8 to PUSHINT256
+			metaLenInBytes = int(scriptArgsOnly[i]) + 1
+		} else {
+			metaLenInBytes = int(scriptArgsOnly[i]) - 11 // PUSHDATA1 to PUSHDATA4
+		}
+		i++
+		var metaLen int
+		//nolint:intrange
+		for j := 0; j < metaLenInBytes; j++ {
+			metaLen += int(scriptArgsOnly[i]) << (8 * j)
```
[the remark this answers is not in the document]
AnnaShaleva: You may use the following compiler test snippet for experiments:
```

func TestByteConstantB(t *testing.T) {
	src := `package foo
	import "github.com/nspcc-dev/neo-go/pkg/interop/convert"
	func Main() int {
		var b = []byte{1}
		x := convert.ToInteger(b)
		return x
	}`
	eval(t, src, big.NewInt(1))
}
```

contracts/proxy/contract.go
```diff
@@ -19,9 +25,14 @@ func OnNEP17Payment(from interop.Hash160, amount int, data any) {
 
 // nolint:deadcode,unused
 func _deploy(data any, isUpdate bool) {
+	args := data.(struct {
```
roman-khimov, line 28: NACK. No parameters, please. You can reuse NNS. I see there is a circular dependency, but you can push hashes into NNS before deploying other contracts (and you can do it in a single transaction).

contracts/proxy/contract.go
roman-khimov: It depends on the overhead, if we can ensure NNS doesn't add much (it's not obvious right now, the code does many more things to resolve) then OK. Just remember that this code is executed for every single transaction and the result is always the same.

This code however is executed concurrently for every transaction...

contracts/container/contract.go
```diff
@@ -195,6 +207,12 @@ func Update(nefFile, manifest []byte, data any) {
 // must contain information about an object placed to a container that was
 // created using [Put] ([PutMeta]) with enabled meta-on-chain option.
 func SubmitObjectPut(metaInformation []byte, sigs [][]interop.Signature) {
+	ctx := storage.GetContext()
+	proxyH := storage.Get(ctx, proxyContractKey).(interop.Hash160)
```
roman-khimov, line 211: ...unlike this one, which is sequential and the cost is _much_ higher here. If we care about breaking the circular dependency I'd leave this getter here and go with NNS for `verify()`.

carpawell: > we care about breaking the circular dependency

Currently it is solved, i do not like how, but it works. i would put NNS everywhere as a simplier solution, but this PR is about optimization so storage getter everywhere now. not sure i understand what current state of this thread? leaving getters?

contracts/container/contract.go
```diff
@@ -112,13 +113,18 @@ func _deploy(data any, isUpdate bool) {
 			}
 		}
 
+		if version < 23_000 {
+			storage.Put(ctx, proxyContractKey, args[5].(interop.Hash160))
```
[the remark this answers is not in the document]
roman-khimov: Just add NNS, there is already code in deploy/update that relies on it. We always have it. Migration tests should reproduce the environment more accurately.

contracts/proxy/contract.go
carpawell: ok

contracts/proxy/contract.go
```diff
@@ -42,8 +53,69 @@ func Update(nefFile, manifest []byte, data any) {
...
+		}
+		i++
+		var metaLen int
+		//nolint:intrange
+		for j := 0; j < metaLenInBytes; j++ {
+			metaLen += int(scriptArgsOnly[i]) << (8 * j)
+			i++
+		}
+
+		metaData = scriptArgsOnly[i : i+metaLen]
```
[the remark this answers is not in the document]
carpawell: hm, is it possible to put smth bad b/w args and contract syscall?

contracts/proxy/contract.go
```diff
@@ -42,8 +53,69 @@ func Update(nefFile, manifest []byte, data any) {
...
+			i++
+			continue
+		}
+		i++
+
+		var metaLenInBytes int
+		if scriptArgsOnly[i] <= 5 { // PUSHINT8 to PUSHINT256
+			metaLenInBytes = int(scriptArgsOnly[i]) + 1
+		} else {
+			metaLenInBytes = int(scriptArgsOnly[i]) - 11 // PUSHDATA1 to PUSHDATA4
```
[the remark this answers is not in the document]
carpawell: hm, `if` part seems redundant, i guess, dropped it. `else` is for packing `[]byte`. is it correct?

contracts/proxy/contract.go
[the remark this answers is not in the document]
carpawell: > Also, I'm not sure about PUSHDATA*, how exactly do you push integers on stack with PUSHDATA* opcodes?

i looked at `emit.AppCall` implementation for byte arrays. method definition is [there](https://github.com/nspcc-dev/neofs-contract/blob/d490c0d5624db991da4166cf7fb347056749b03b/contracts/container/contract.go#L197) and i am parsing `metaInformation []byte` in this line

> I'd expect that convert.ToInteger should do the trick

hm, thought it is smth more clever, but this is just a `.(int)` converter, i tried it earlier (and tried `convert.ToInteger` too) and it did not work. or i did smth not right: 

```
func TestByteStringIntegerConversion(t *testing.T) {
	src := `package foo
	import "github.com/nspcc-dev/neo-go/pkg/interop/convert"
	func Main() int {
		var b = []byte{0, 17}
		x := convert.ToInteger(b)
		return x
	}`
	eval(t, src, 17)
}
```

got: 
```
        	Error:      	Not equal: 
        	            	expected: int(17)
        	            	actual  : *big.Int(4352)
```

contracts/proxy/contract.go
```diff
@@ -42,8 +53,69 @@ func Update(nefFile, manifest []byte, data any) {
...
+	if string(scriptNoContract[scriptLen-len(expMethod):]) != expMethod {
+		return common.ContainsAlphabetWitness()
+	}
+
+	scriptArgsOnly := scriptNoContract[:scriptLen-len(expMethod)-1-1]
+	scriptLen = len(scriptArgsOnly)
+	var sigs []interop.Signature
+	var metaData []byte
+	for i := 0; i < scriptLen; {
+		if scriptArgsOnly[i] == 12 { // PUSHDATA1
```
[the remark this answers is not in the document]
carpawell: my idea was to write as little code as possible. also i think it is still valid since we know for sure that this is a Container's `SubmitObjectPut`. even if it is not ok here it will fail further in the Container's execution. happy path is checked and i expect most of the calls to be ok. attackers will fail anyway

contracts/proxy/contract.go
```diff
@@ -42,8 +74,67 @@ func Update(nefFile, manifest []byte, data any) {
...
+	if string(tx.Script[len(tx.Script)-5:]) != string([]byte{65, 98, 125, 91, 82}) { // SystemContractCall
+		return common.ContainsAlphabetWitness()
+	}
+	scriptNoSyscall := tx.Script[:len(tx.Script)-5]
+	scriptLen := len(scriptNoSyscall)
+	if !cnrH.Equals(scriptNoSyscall[scriptLen-20:]) {
+		return common.ContainsAlphabetWitness()
+	}
+
+	scriptNoContract := scriptNoSyscall[:scriptLen-20-1-1]
```
roman-khimov, line 91: PUSHDATA1 + length? These bytes need to be checked as well.

contracts/proxy/contract.go
```diff
@@ -42,8 +74,67 @@ func Update(nefFile, manifest []byte, data any) {
...
+	scriptLen = len(scriptNoContract)
+	expMethod := "submitObjectPut"
+	if scriptLen < len(expMethod) {
+		return common.ContainsAlphabetWitness()
+	}
+	if string(scriptNoContract[scriptLen-len(expMethod):]) != expMethod {
+		return common.ContainsAlphabetWitness()
+	}
+
+	scriptArgsOnly := scriptNoContract[:scriptLen-len(expMethod)-1-1]
```
roman-khimov, line 101: Frankly, this backward reading is not very friendly, iterating in natural order is easier to understand.

contracts/proxy/contract.go
```diff
@@ -42,8 +53,69 @@ func Update(nefFile, manifest []byte, data any) {
...
+	scriptLen = len(scriptArgsOnly)
+	var sigs []interop.Signature
+	var metaData []byte
+	for i := 0; i < scriptLen; {
+		if scriptArgsOnly[i] == 12 { // PUSHDATA1
+			newI := i + 2 + interop.SignatureLen
+			sigs = append(sigs, scriptArgsOnly[i+2:newI])
+			i = newI
```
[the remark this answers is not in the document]
roman-khimov: Size check is mandatory, yes. Although invalid signature or non-signature will fail verification.

deploy/contracts.go
```diff
@@ -334,51 +339,54 @@ func syncNeoFSContract(ctx context.Context, prm syncNeoFSContractPrm) (util.Uint
...
-				continue
-			}
-
-			missingDomainRecord = errors.Is(err, errMissingDomain) || errors.Is(err, errMissingDomainRecord)
-			if !missingDomainRecord {
-				if errors.Is(err, errInvalidContractDomainRecord) {
-					l.Error("contract's domain record is invalid/unsupported, will wait for a background fix", zap.Error(err))
-				} else {
-					l.Error("failed to read on-chain state of the contract record by NNS domain name, will try again later", zap.Error(err))
+		if !nnsRecordWithoutContractIsOK {
```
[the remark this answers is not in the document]
roman-khimov: > error	deploy/contracts.go:425	failed to send Notary request deploying the contract, will try again later	{"contract": "NeoFS Container", "domain": "container.neofs", "error": "script failed (FAULT state) due to an error: at instruction 8 (SYSCALL): contract already exists"}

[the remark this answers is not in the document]
carpawell: > Not entirely sure what's the consequence of this in case of some problems with contracts.

changes were required since it was not expected to have a record in NNS and still have a need to deploy a contract

contracts/proxy/contract.go
carpawell: ok, but still: https://github.com/nspcc-dev/neofs-contract/pull/494#discussion_r2129432493. i cannot understand where i am wrong

contracts/proxy/contract.go
```diff
@@ -42,8 +74,67 @@ func Update(nefFile, manifest []byte, data any) {
...
+	}
+	scriptNoSyscall := tx.Script[:len(tx.Script)-5]
+	scriptLen := len(scriptNoSyscall)
+	if !cnrH.Equals(scriptNoSyscall[scriptLen-20:]) {
+		return common.ContainsAlphabetWitness()
+	}
+
+	scriptNoContract := scriptNoSyscall[:scriptLen-20-1-1]
+	scriptLen = len(scriptNoContract)
+	expMethod := "submitObjectPut"
```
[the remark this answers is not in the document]
carpawell: ok

contracts/proxy/contract.go
```diff
@@ -42,8 +74,67 @@ func Update(nefFile, manifest []byte, data any) {
...
+	if scriptLen < len(expMethod) {
```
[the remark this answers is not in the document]
carpawell: dropped because of some of the threads above. initially, it was possible to have a correct contract call, which was not a "submitObjectPut," so i checked the length. now, at this line, only a single method is expected

contracts/proxy/contract.go
carpawell: i feel like a fast fail if it is not a required contract/method/contract syscall was a better choice. also i feel this order is more natural ("i call this contract, this method...")

contracts/proxy/contract.go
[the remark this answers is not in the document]
roman-khimov: It'd be the same `PUSHDATA1` for metadata as for signatures in this case. And it can overflow 255 in which case it'd be `PUSHDATA2` (but it's a different problem).

roman-khimov (approved):
Looks legit.

contracts/proxy/contract.go
```diff
@@ -40,10 +73,128 @@ func Update(nefFile, manifest []byte, data any) {
...
+		metaLenInBytes = 2
+	case 0x0E: // PUSHDATA4
+		metaLenInBytes = 4
+	default:
+		panic("invalid meta push opcode " + std.Itoa(op, 16))
+	}
+	var metaLen int
+	//nolint:intrange
+	for j := 0; j < metaLenInBytes; j++ {
+		metaLen += int(script[i]) << (8 * j)
```
carpawell, line 148: and one more non-stylish push: i still think that is the only way to parse byte array length, `convert.ToInteger` works differently, @AnnaShaleva, @roman-khimov, validate, please

roman-khimov (approved):

PR #507, opened 2025-08-14
container: calculate container estimations on contract side

### LINKED ISSUE #501: Accept container estimations via contract method
I'm always frustrated when we use unreliable protocol for container size estimations that is also missing some useful data.

Add a method to `container` contract to accept and store size and object count estimations from nodes. It's to be invoked by nodes with some limit (like no more than three times per epoch). We need to always have current size/count estimation for container based on submitted data and some epoch average for subsequent basic income settlement. Meta-enabled case is currently out of scope (we can have exact data there). Provide methods to get current state and previous epoch data.

carpawell:
Accept container estimations report made by storage nodes directly from every netmap member (limiting max number of reports per epoch by 3). Keep an average value for every container. Add API for fetching new estimations and allow checking what exact report every storage node made. Extend "container estimation" term to (size, objectsNumber) pair. Closes #501.


cthulhu-rider (commented):
looks correct

PR #511, opened 2025-09-02
container: move deprecation notices into proper places

roman-khimov:
38ac6f61f346d6c9e4a9c9d70e1d5e1b01b1845d breaks all Go conventions for function comments.


roman-khimov:
It's simple, function comment always starts with the function name. Not with "Deprecated".

PR #514, opened 2025-09-02
Feat/space quotas

carpawell:


contracts/container/contract.go
```diff
@@ -1298,6 +1307,139 @@ func NewEpoch(epochNum int) {
...
+	if v == nil {
+		return Quota{}
+	}
+
+	q := std.Deserialize(v.([]byte))
+
+	return q.(Quota)
+}
+
+// PutUserQuota sets size quota that limits all space used for storing objects
```
cthulhu-rider, line 1379: this can also be used out of container ownership context. For example, if I have no containers, but I upload data to someone else's. Right? lookin at implementation

contracts/container/contract.go
```diff
@@ -1298,6 +1307,139 @@ func NewEpoch(epochNum int) {
...
+	storage.Put(ctx, key, std.Serialize(Quota{Size: size, Hard: hard}))
+	runtime.Notify("UserQuotaSet", user, size, hard)
+}
+
+// UserQuota returns user size quota for every container he owns. If no
+// limitation has been set, returns empty value (limitation is set to 0;
+// hard is false). Non-positive limitation must be treated as no limits.
+// Panics if user is incorrect.
+func UserQuota(user []byte) Quota {
+	if len(user) != 25 {
```
cthulhu-rider, line 1426: consider sharing const in `common` package

roman-khimov (commented):
Can we have _both_ soft and hard limits set for a container? Do we track space occupied by account?

contracts/container/contract.go
```diff
@@ -1297,6 +1307,139 @@ func NewEpoch(epochNum int) {
...
+
+	storage.Put(ctx, key, std.Serialize(Quota{Size: size, Hard: hard}))
+	runtime.Notify("ContainerQuotaSet", cID, size, hard)
+}
+
+// ContainerQuota returns container size quota set by cID's owner. If no
+// limitation has been set, returns empty value (limitation is set to 0;
+// hard is false). Non-positive limitation must be treated as no limits.
+// Panics if cID is incorrect or container does not exist.
+func ContainerQuota(cID interop.Hash256) Quota {
```
AnnaShaleva, line 1357: This method should be marked as "safe" in the contract's manifest.

contracts/container/contract.go
```diff
@@ -1297,6 +1307,139 @@ func NewEpoch(epochNum int) {
...
+
+	storage.Put(ctx, key, std.Serialize(Quota{Size: size, Hard: hard}))
+	runtime.Notify("UserQuotaSet", user, size, hard)
+}
+
+// UserQuota returns user size quota for every container he owns. If no
+// limitation has been set, returns empty value (limitation is set to 0;
+// hard is false). Non-positive limitation must be treated as no limits.
+// Panics if user is incorrect.
+func UserQuota(user []byte) Quota {
```
AnnaShaleva, line 1425: Ditto.

contracts/container/contract.go
```diff
@@ -1297,6 +1307,139 @@ func NewEpoch(epochNum int) {
...
+			r := iterator.Value(it).(NodeReport)
+			currentSize += r.ContainerSize
+		}
+		if currentSize >= size {
+			panic(cst.ErrorQuotaLimitAlreadyExceeded)
+		}
+	}
+
+	key := []byte{containerQuotaKeyPrefix}
+	key = append(key, cID...)
```
AnnaShaleva, line 1347: Usually we move such code to helpers like `getContainerQuotaKey` and reuse from getter/setter, but it's up to you.

contracts/container/contract.go
```diff
@@ -1298,6 +1307,139 @@ func NewEpoch(epochNum int) {
...
+	currEpoch := contract.Call(netmapContractAddr, "epoch", contract.ReadOnly).(int)
+
+	if size > 0 {
+		var currentSize int
+		it := IterateReports(currEpoch, cID)
+		for iterator.Next(it) {
+			r := iterator.Value(it).(NodeReport)
+			currentSize += r.ContainerSize
+		}
+		if currentSize >= size {
```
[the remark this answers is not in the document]
carpawell: the check was dropped

contracts/container/contract.go
```diff
@@ -1297,6 +1307,139 @@ func NewEpoch(epochNum int) {
...
+			panic(cst.ErrorQuotaLimitAlreadyExceeded)
```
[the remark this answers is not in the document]
carpawell: ok

contracts/container/contract.go
carpawell: tbh initially i meant that user quota == sum of every object that was put to a container this user has. i did not think about sum of objects that the user uploaded (even to other users' containers). i am not sure it can be precisely calculated how many objects a certain user has uploaded by a node, it just does not know the other node's infos. moreover,  i do not understand what user's quota means then: _this_ user set it and _this_ user uploads objects with _this_ user's wallet. what is the restriction then?

@roman-khimov

contracts/container/contract.go
```diff
@@ -1298,6 +1307,139 @@ func NewEpoch(epochNum int) {
...
+// objects in cID (including object replicas). Non-positive size sets no
+// limitation. if hard is true, storage nodes decline any exceeding object at
+// PUT operation; if false, nodes are instructed to warn only, without denial
+// of service. Call must be signed by cID's owner. Limit can be changed with
+// repeated call.
+// Panics if cID is incorrect, container does not exist or new limitation already
+// exceeded (according to registered NodeReports).
+func PutContainerQuota(cID interop.Hash256, size int, hard bool) {
+	if len(cID) != interop.Hash256Len {
+		panic("wrong container id")
```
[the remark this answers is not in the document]
carpawell: it is not the first time we discuss this exact panic, the contract even before this PR had both versions, and i do not understand why one of them is preferable (or not) over the other., fixed

roman-khimov:
> > Can we have both soft and hard limits set for a container?
> 
> No.

I think @EESergey might like it.

EESergey:
> > > Can we have both soft and hard limits set for a container?
> > 
> > 
> > No.
> 
> I think @EESergey might like it.

The soft quota is for notification.
Stopping the upload of objects when the hard quota is reached can be an emergency.
There should be a notification mechanism before the upload stops.

contracts/container/contract.go
cthulhu-rider: doubt this fact. If so, to upload data, any side would need to create a container

roman-khimov (approved):

PR #517, opened 2025-09-12
Fix/No epoch in size reports

### LINKED ISSUE #516: Container estimations are tied to epochs
We still accept/store estimations per-epoch. This is not what #501 tried to convey and this doesn't allow for https://github.com/nspcc-dev/neofs-node/issues/3543#issuecomment-3240144475.

carpawell:
Closes #516.


contracts/container/contract.go
```diff
@@ -911,6 +914,43 @@ func CommitContainerListUpdate(cID interop.Hash256, replicas []uint8) {
 		}
 	}
 
+	// update reports stat after netmap change
```
roman-khimov, line 917: Exactly the thing I want to avoid. Epoch tick should not trigger massive changes in contracts.

roman-khimov (commented):
1. Always keep track of current sum values, you don't need epoch ticks for that.
2. Keep track of current epoch average, store previous epoch average, but don't do anything to them per-epoch, that's the trick.

contracts/container/contract.go
carpawell: @roman-khimov, this is not an epoch tick; this is a netmap change, which is done only when some storage nodes leave it. i see no other possibility to drop unused reported values. the previous approach was OK since reports were updated every epoch (an epoch was included in the report's storage key), and left nodes are not included naturally

roman-khimov: OK, makes some sense.

contracts/container/contract.go
```diff
@@ -1489,13 +1478,31 @@ func addContainer(ctx storage.Context, id, owner, container []byte) {
...
+
+	summaryKey := append([]byte{reportsSummary}, id...)
+	summaryRaw := storage.Get(ctx, summaryKey)
+	if summaryRaw != nil {
+		rSummary := std.Deserialize(summaryRaw.([]byte)).(NodeReportSummary)
+		diff := -rSummary.ContainerSize
+
+		totalUserKey := userTotalStorageKey(owner)
+		oldVal := storage.Get(ctx, totalUserKey).(int)
+		storage.Put(ctx, totalUserKey, oldVal+diff)
```
roman-khimov, line 1491: Should be deleted if zero.

roman-khimov (commented):
Seems to be good enough for current value tracking. But how about the previous epoch average for billing?

contracts/container/contract.go
```diff
@@ -1328,14 +1322,9 @@ func IterateAllContainerSizes(epoch int) iterator.Iterator {
 	return storage.Find(ctx, key, storage.RemovePrefix|storage.DeserializeValues)
 }
 
-// NewEpoch method removes all container size estimations from epoch older than
-// epochNum + 3. It can be invoked only by NewEpoch method of the Netmap contract.
+// NewEpoch method does nothing.
```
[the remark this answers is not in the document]
carpawell: lets try to imagine we gonna make it nice: check new commits, please

carpawell:
> But how about the previous epoch average for billing?

@roman-khimov, isnt checking actual values and paying for it enough?

roman-khimov:
Nope. We can leave it for another issue, but "actual" values are incorrect for proper billing.

PR #535, opened 2025-11-07
deploy: allow calling other contracts in update process

carpawell:


deploy/contracts.go
```diff
@@ -228,7 +228,15 @@ func syncNeoFSContract(ctx context.Context, prm syncNeoFSContractPrm) (util.Uint
 
 	initProxyCommitteeActor := func(proxyContract util.Uint160) error {
 		var err error
-		proxyCommitteeActor, err = newProxyCommitteeNotaryActor(prm.blockchain, prm.localAcc, prm.committee, proxyContract)
+
+		if len(prm.validatorsDeployAllowedContracts) != 0 {
+			proxyCommitteeActor, err = _newCustomCommitteeNotaryActor(prm.blockchain, prm.localAcc, prm.committee, notary.FakeContractAccount(proxyContract), func(signer *transaction.Signer) {
+				signer.Scopes = transaction.CalledByEntry | transaction.CustomContracts
+				signer.AllowedContracts = prm.validatorsDeployAllowedContracts
```
AnnaShaleva: If you need more strict constraints, then as Roman suggested, you may use Rules scope:
https://github.com/nspcc-dev/neo-go/blob/8385780f11cdb5df9e8e9279d4f651fa2b8344c6/pkg/core/transaction/witness_scope.go#L27
Check the "Rules" part in https://neospcc.medium.com/thou-shalt-check-their-witnesses-485d2bf8375d.

AnnaShaleva: I suppose so. Just note, you should add `ALLOW` action to this condition (the default one is `DENY`).

roman-khimov: Consider third-party `deploy` users, currently they're restricted and can't do actions in FS contracts that easy. This behavior should be kept, only FS contract updates can do things in FS contracts.

carpawell: ok, will change. but, tbh, still do not understand how this can be misused: almost everywhere we have alphabet signature checks; our contracts can be called only by our contracts

deploy/contracts.go
```diff
@@ -468,10 +488,17 @@ func syncNeoFSContract(ctx context.Context, prm syncNeoFSContractPrm) (util.Uint
...
 			}
 		} else {
-			if prm.isProxy && proxyCommitteeActor == nil {
-				err = initProxyCommitteeActor(onChainState.Hash)
-				if err != nil {
-					return util.Uint160{}, err
+			if proxyCommitteeActor == nil {
+				if prm.isProxy {
+					err = initProxyCommitteeActor(onChainState.Hash, onChainState.Hash)
+					if err != nil {
```
roman-khimov: Too many duplicating lines.

deploy/contracts.go
```diff
@@ -226,24 +226,44 @@ func syncNeoFSContract(ctx context.Context, prm syncNeoFSContractPrm) (util.Uint
...
 
-	initProxyCommitteeActor := func(proxyContract util.Uint160) error {
+	initProxyCommitteeActor := func(proxyContract, deployedContract util.Uint160) error {
 		var err error
-		proxyCommitteeActor, err = newProxyCommitteeNotaryActor(prm.blockchain, prm.localAcc, prm.committee, proxyContract)
+
+		if len(prm.validatorsDeployAllowedContracts) != 0 {
+			var (
+				scriptHashConditions      transaction.ConditionOr
+				calledByContractCondition = transaction.ConditionCalledByContract(deployedContract)
```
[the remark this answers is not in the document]
carpawell: it's like deploy callers should never fill `prm.validatorsDeployAllowedContracts`. i expect any caller that has all necessary alphabet signatures to be a sane app. on the other hand, if you _need_ alpha signature inside other contracts, for your custom contract, then what else can we do? we just allow it, i see no other opportunities

PR #549, opened 2025-11-20
NEP-11 containers

cthulhu-rider:
* based on and blocked by #534

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -57,7 +57,7 @@ jobs:
           cache: true
 
       - name: Write coverage profile
-        run: go test -v ./... -coverprofile=./coverage.txt -coverpkg=./contracts/...
+        run: go test -v ./... -coverprofile=./coverage.txt -coverpkg=./contracts/... -timeout 20m
 
       - name: Upload coverage results to Codecov
         uses: codecov/codecov-action@v4


contracts/container/contract.go
```diff
@@ -1721,6 +1722,66 @@ func OwnerOf(tokenID []byte) interop.Hash160 {
...
+// Tokens implements optional NEP-11 method.
+func Tokens() iterator.Iterator {
+	return storage.Find(storage.GetReadOnlyContext(), []byte{containerKeyPrefix}, storage.KeysOnly|storage.RemovePrefix)
+}
+
+// Properties returns properties of referenced container.
+//
+// The properties are:
+//
+//	name: Base58-encoded tokenID;
```
roman-khimov: It can be specified in attributes as well.

contracts/container/contract.go
```diff
@@ -1721,6 +1722,66 @@ func OwnerOf(tokenID []byte) interop.Hash160 {
...
+//	owner: Base64-encoded [Info.Owner];
+//	nonce: Base64-encoded [Info.Nonce];
+//	basicACL: [Info.BasicACL];
+//	storagePolicy: Base64-encoded [Info.StoragePolicy];
```
roman-khimov: These two are somewhat questionable to me as well. Not that it's an error to return them, but we do create some potential naming conflict. We can try concentrating on attributes first.

contracts/container/contract.go
```diff
@@ -1721,6 +1722,66 @@ func OwnerOf(tokenID []byte) interop.Hash160 {
...
+		if storage.Get(ctx, append([]byte{deletedKeyPrefix}, tokenID...)) != nil {
+			panic(cst.ErrorDeleted)
+		}
+
+		panic(cst.NotFoundError)
+	}
+	bin := binItem.([]byte)
+
+	var cnr Info
+	if item := storage.Get(ctx, append([]byte{infoPrefix}, tokenID...)); item == nil {
```
roman-khimov: Prioritize new format.

roman-khimov (commented):
We need to support this as well for proper event tracking:

> A token contract which creates new tokens MUST trigger a Transfer event with the from address set to null when tokens are created.
> A token contract which burns tokens MUST trigger a Transfer event with the to address set to null when tokens are burned. 

It's creating and deleting containers basically.

contracts/container/contract.go
cthulhu-rider: so?

roman-khimov: I'd use `Name` attribute in this case. Per NEP-11 `name` must be provided and that's exactly what `Name` attribute is. But the attribute is optional, so a fallback to token id is fine.

contracts/container/contract.go
```diff
@@ -1563,6 +1559,168 @@ func Version() int {
 	return common.Version
 }
 
+// Symbol returns static 'CONTAINER'.
+//
+// Symbol returns implements NEP-11 method.
```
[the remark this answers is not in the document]
cthulhu-rider: fixed

contracts/container/contract.go
```diff
@@ -1563,6 +1559,168 @@ func Version() int {
...
+func Symbol() string {
+	return "CONTAINER"
```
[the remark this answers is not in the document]
cthulhu-rider: done

contracts/container/contract.go
```diff
@@ -1563,6 +1559,168 @@ func Version() int {
...
+	if item := storage.Get(ctx, key); item == nil {
+		cnr = fromBytes(bin)
+	} else {
+		cnr = std.Deserialize(item.([]byte)).(Info)
+	}
+
+	if !runtime.CheckWitness(cnr.Owner) {
+		return false
+	}
+
```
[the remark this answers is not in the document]
cthulhu-rider: sensible, done

contracts/container/contract.go
```diff
@@ -1721,6 +1722,66 @@ func OwnerOf(tokenID []byte) interop.Hash160 {
...
+func Tokens() iterator.Iterator {
+	return storage.Find(storage.GetReadOnlyContext(), []byte{containerKeyPrefix}, storage.KeysOnly|storage.RemovePrefix)
+}
+
+// Properties returns properties of referenced container.
+//
+// The properties are:
+//
+//	name: Base58-encoded tokenID;
+//	owner: Base64-encoded [Info.Owner];
```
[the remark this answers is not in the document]
cthulhu-rider: dropped

contracts/container/contract.go
cthulhu-rider: dropped for now

roman-khimov (dismissed):
What happened to coverage test here?

roman-khimov:
Usually this happens when some test gets stuck because locally they all run much faster than 10m. But if it's valid time spent here then OK.

PR #558, opened 2025-11-28
Support container deletion lock

cthulhu-rider:
also wanna w8 4 #555 and provide option to set lock attribute the same way


contracts/container/contract.go
```diff
@@ -721,6 +733,10 @@ func Delete(containerID []byte, signature interop.Signature, token []byte) {
 // container does not exist. On success, Remove throws 'Removed' notification
 // event.
 //
+// If the container has '__NEOFS__LOCK_UNTIL' attribute with timestamp that has
+// not passed yet, Remove throws exception containing [cst.ErrorLocked]. The
+// exception includes both current and lock timestamps in Unit Timestamp format.
```
roman-khimov: Unix

contracts/container/contract.go
```diff
@@ -1941,6 +1968,21 @@ func scriptHashToAddress(h interop.Hash160) []byte {
 	return addr
 }
 
+func checkLock(cnr Info) string {
+	for i := range cnr.Attributes {
+		if cnr.Attributes[i].Key == attributeLock {
+			// TODO: (issue) throws 'at instruction 1 (SYSCALL): invalid format' exception. Catch and return clear one instead.
```
roman-khimov: The biggest problem is that this makes "__NEOFS__LOCK_UNTIL=randomthings" forever undeleteable. For deletion this problem should just be ignored (incorrect attribute -> lock doesn't work), but this requires a check before `Atoi10` call. This attribute should also be checked at container creation, we can forbid such incorrectly configured containers right away. In that case removal code can rely on storage consistency, attribute should be correct there and not requiring additional checks. Transaction failure is also OK in this case, no container created -- no problem (even though it's hard to understand what's the real cause).

cthulhu-rider: creation check is required ofc. Currently, all such checks are performed by IR. At the same time, we can easily do this by contract now. Lets start this from lock?

PR #591, opened 2026-03-25
deploy/netmap: deprecate homomorphic

carpawell:
Make it impossible to deploy networks with homomorphic settings tuned: it should be always disabled for now. Refs #3847.


roman-khimov (commented):
Consider #590.

roman-khimov (approved):

PR #601, opened 2026-07-07
Proxy with none scope for alpha

roman-khimov:


carpawell (commented):
Why does it make things simpler/more secure/better, etc? Isn't witness scope made for such cases?

roman-khimov:
Verify is called in verification context, it's used to ensure proxy contract witness for GAS spending or other purposes allowed by the respective scope. It relies on alphabet signature which is what we want for any spends from proxy contract account. At the moment it also requires this signature to have at least CalledByEntry scope which is not really relevant for verification context (it's read-only) but it gives additional permissions for application context (in which the transaction script is executed) and these permissions can't be avoided. Some transactions naturally want to give this permission (like epoch tick, it needs alpha witness anyway), but some may want to avoid it if there are no actions done that require alpha witness. This is what this PR allows to happen, now transaction can be created with None scope for both proxy and alpha witnesses.

carpawell (approved):

PR #608, opened 2026-07-14
alphabet: tune emission rules, leave more to proxy

roman-khimov:
IR doesn't use that much of GAS, all of real actions are performed via Notary, so current emission rate is obviously excessive. Mainnet at the same time spends quite a lot of GAS from Proxy for its 300+ containers, so we better have more left for proxy rather than IR.


carpawell (approved):
Magic numbers to me, but ok.

PR #612, opened 2026-08-05
netmap: make network map versioned

carpawell:
I am not sure we are ready to detach network map from epochs fully. Also, there is no "Merkle root hash" suggestion from the issue, I did not understand it.
~Fast new epoch after first nodes are entered TBD.~ Done.


contracts/netmap/config.yml
```diff
@@ -34,6 +34,10 @@ events:
     parameters:
     - name: contract
       type: Hash160
+  - name: NewNetmapVersion
```
roman-khimov: Can just be `NewNetmap`.

contracts/netmap/contract.go
```diff
@@ -620,6 +640,40 @@ func fillNetmap(epoch int) {
 			storage.LocalPut(append(epochPrefix, kv.Key...), std.Serialize(n2))
 		}
 	}
+
+	if !netmapChanged {
+		// there are no "lost" nodes, and no new nodes, still there could be
+		// manually removed from candidates nodes
```
roman-khimov: All of these changes (but lost nodes that do require some looping now even though can be improved in future) can be tracked beforehand, not requiring a number of whole-netmap loops (netmaps can be big), any state-changing action via the netmap contract can trigger a flag in the same contract. Then it's just checking a flag and that's it.

contracts/netmap/contract.go
```diff
@@ -600,7 +613,9 @@ func fillNetmap(epoch int) {
 	var (
 		cleanupThreshold = CleanupThreshold()
 		epochPrefix      = append([]byte(node2NetmapPrefix), convert.Uint32ToBytesBE(uint32(epoch))...)
```
roman-khimov: Wasting additional storage on this is also counterproductive. Even for transition period I'd prefer some epoch to version map, not a full copy of it.

carpawell: do you mean completely reworking the storage scheme? store netmap by versions, not by epochs?

roman-khimov: Yes.

contracts/netmap/contract.go
carpawell: can be done, but it may lead to false positive events (when someone drops their node and then adds it in the same epoch). though it is not a problem to me. also, can be fixed later with netmap hashing

contracts/netmap/contract.go
carpawell: reworked, can you check?

tests/netmap_test.go
```diff
@@ -312,8 +371,6 @@ func TestAddNode(t *testing.T) {
 	}
 	// Current map is still empty.
 	checkZeroList("listNodes")
-	// And historic data is gone.
-	checkZeroList("listNodes", 1)
```
carpawell, line 374: had to drop it since the threshold is smth different now and the contract _may_ have more versions of netmap than the epoch threshold says to have. maybe we need to rework this threshold too, it would not be difficult, but it will change the contract API and its meaning

contracts/netmap/contract.go
```diff
@@ -223,6 +235,23 @@ func updateCandidateState(publicKey interop.PublicKey, state nodestate.Type) {
 	runtime.Notify("UpdateStateSuccess", publicKey, state)
 }
 
+func setNetmapChangedFlag(flag bool) {
+	var v = byte(0)
+	if flag {
```
roman-khimov: https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/interop@v0.0.0-20260807115931-cfee8827ddfd/convert#ToBytes

contracts/netmap/contract.go
```diff
@@ -360,8 +389,24 @@ func IsStorageNodeInEpoch(key interop.PublicKey, epoch int) bool {
 	return getStorageNodeInEpoch(key, epoch) != nil
 }
 
+func nmVersionForEpoch(epoch int) uint32 {
+	var (
+		it          = storage.LocalFind([]byte{epochToNetmapVersionKey}, storage.RemovePrefix)
+		verForEpoch int
+	)
+	for iterator.Next(it) {
```
roman-khimov, line 404: Refs. https://github.com/neo-project/neo/issues/4039

contracts/netmap/contract.go
```diff
@@ -580,6 +631,8 @@ func GetEpochBlockByTime(ts int) int {
 }
 
 func removeFromNetmap(key interop.PublicKey) {
+	runtime.Log("removing from map?")
```
roman-khimov: DEbug

contracts/netmap/contract.go
```diff
@@ -609,23 +665,87 @@ func fillNetmap(epoch int) {
...
+func incNetmapVersion() uint32 {
+	var (
+		k      = []byte{netmapVersionKey}
+		ver    uint32
+		verRaw = storage.LocalGet(k)
+	)
+	if verRaw == nil {
+		ver = 0
+	} else {
+		ver = convert.BytesBEToUint32(verRaw)
```
roman-khimov: Why fixed-length format for version? It's somewhat less effective. Usually it's used for _keys_ where variable length (and LE) is problematic, but it's a value here.

contracts/netmap/contract.go
```diff
@@ -609,23 +665,87 @@ func fillNetmap(epoch int) {
...
+	}
+	var nmIt = storage.LocalFind([]byte(node2NetmapPrefix), storage.KeysOnly)
+	for iterator.Next(nmIt) {
+		ver := convert.BytesBEToUint32(iterator.Value(nmIt).([]byte)[1:5])
+		if ver >= latestValidVersion {
+			return
+		}
+
+		storage.LocalDelete(iterator.Value(nmIt).([]byte))
 	}
```
roman-khimov, line 730: How about the epoch-version storage? It shouldn't store stale entries.

contracts/netmap/contract.go
```diff
@@ -609,23 +665,87 @@ func fillNetmap(epoch int) {
...
+	} else {
+		ver = convert.BytesBEToUint32(verRaw)
 	}
+
+	ver++
+	storage.LocalPut(k, convert.Uint32ToBytesBE(ver))
+	return ver
 }
 
 func dropNetmap(epoch int) {
```
roman-khimov: We can also store a number of _versions_ in the contract, not a number of epochs. Same storage, better functionality.

contracts/netmap/contract.go
```diff
@@ -642,3 +762,26 @@ func cleanup(epoch int) {
...
+	var (
+		newVersion = incNetmapVersion()
+		currEpoch  = Epoch()
+		nmIt       = storage.LocalFind([]byte(node2NetmapPrefix), storage.None)
+	)
+
+	storage.LocalPut(append([]byte{epochToNetmapVersionKey}, convert.Uint32ToBytesBE(uint32(currEpoch))...), convert.ToBytes(newVersion))
+	storage.LocalPut([]byte{netmapChangedFlagKey}, []byte{0})
+
+	for iterator.Next(nmIt) {
```
roman-khimov, line 766: You have a number of maps stored, all of them should be migrated, not just the latest.

contracts/netmap/config.yml
```diff
@@ -34,6 +34,10 @@ events:
     parameters:
     - name: contract
       type: Hash160
+  - name: NewNetmap
+    parameters:
+      - name: version
+        type: Integer
```
roman-khimov, line 40: You have new safe methods as well. And you probably need more of them (versioned netmap getters).

contracts/netmap/contract.go
```diff
@@ -200,6 +200,12 @@ func AddNode(n Node2) {
 	// new network map version without changes
 	setNetmapChangedFlag(true)
 	runtime.Notify("AddNode", n.Key, n.Addresses, n.Attributes)
+
+	numOfContainers := contract.Call(common.ResolveFSContract("container"), "totalSupply", contract.ReadOnly).(int)
```
roman-khimov, line 204: Looks like it's not related to versioning and can be merged separately (more quickly).

contracts/netmap/contract.go
carpawell: dedebuged

contracts/netmap/contract.go
carpawell: no practical use cases, made by inertia, fixed

contracts/netmap/contract.go
carpawell: agree, fixed

contracts/netmap/contract.go
carpawell: do you mean this: https://github.com/nspcc-dev/neofs-contract/pull/612#discussion_r3776639630?

contracts/netmap/contract.go
carpawell: do you mind a one-time history drop on migration? it would simplify migration, and it would not lose important info, imo

contracts/netmap/contract.go
carpawell: https://github.com/nspcc-dev/neofs-contract/pull/613

contracts/netmap/contract.go
roman-khimov: I fear some SN code may freak out on `epoch-1` map unavailability. Older ones are not that interesting in general, but penultimate one is requested a lot.

contracts/netmap/contract.go
roman-khimov: Yeah, the same issue of threshold/history applicability to the current scheme. Storage-wise it costs the same to have a number of versions, while epoch span would be wider and it's OK to me.

contracts/netmap/contract.go
carpawell: fixed migration and added test for calling the previous epoch

contracts/netmap/contract.go
```diff
@@ -367,8 +400,24 @@ func IsStorageNodeInEpoch(key interop.PublicKey, epoch int) bool {
...
 
+func nmVersionForEpoch(epoch int) uint32 {
+	var (
+		it          = storage.LocalFind([]byte{epochToNetmapVersionKey}, storage.RemovePrefix)
+		verForEpoch int
+	)
+	for iterator.Next(it) {
+		kv := iterator.Value(it).(storage.KeyValue)
+		verForEpoch = convert.ToInteger(kv.Value)
+		if convert.BytesBEToUint32(kv.Key) >= uint32(epoch) {
```
roman-khimov: Why `>=`? Say you have version 1 at 100, version 2 at 110 and requesting a map for height 105. The valid one is version 1, but you'll return version 2. You can try backwards iteration to fix it in absence of https://github.com/neo-project/neo/issues/4039

contracts/netmap/contract.go
```diff
@@ -616,23 +674,93 @@ func fillNetmap(epoch int) {
...
+
+	newVersion = incNetmapVersion()
+	storage.LocalPut(append([]byte{epochToNetmapVersionKey}, convert.Uint32ToBytesBE(uint32(epoch))...), convert.ToBytes(newVersion))
+	setNetmapChangedFlag(false)
+	runtime.Notify("NewNetmap", newVersion)
+}
+
+func netmapVersion() int {
+	vRaw := storage.LocalGet([]byte{netmapVersionKey})
+	if vRaw == nil {
```
roman-khimov: Can't happen.

contracts/netmap/contract.go
```diff
@@ -535,6 +584,12 @@ func UnusedCandidate() Candidate {
 	return Candidate{}
 }
 
+// NetworkMapVersion return current network map version. Version monotonically
+// increases with each change to the network map.
+func NetworkMapVersion() int {
+	return netmapVersion()
```
roman-khimov: Can internalize it then and call `NetworkMapVersion` in other places. Each call has some cost to it.

contracts/netmap/contract.go
```diff
@@ -649,3 +777,74 @@ func cleanup(epoch int) {
...
+		}
+
+		newKey = append(newKey, kv.Key[1+4:]...)
+
+		storage.LocalPut(newKey, kv.Value)
+		storage.LocalDelete(kv.Key)
+	}
+
+	if prevEpochNodes != currEpochNodes {
+		// maps are different, keep both version
```
roman-khimov: You can keep both in any event, I doubt it matters much for migration, but makes things simpler at the same time. Eventually old map is to be deleted. Also, version 0 is not the best for a real map (much like epoch 0 never has any nodes version 0 can't really have any either for new networks).

carpawell: what cons of this check, GAS? version 0 will be available only for rare cases when we need to support previous epoches for outdated API

roman-khimov: It's just easier to keep two versions and not think much about it.

contracts/netmap/contract.go
carpawell: really that matter? fixed

contracts/netmap/contract.go
carpawell: hm, fixed and added a test for it based on both epoch and version counters

contracts/netmap/contract.go
carpawell: well, ok, but it was already done, and it stored the different netmaps only...

roman-khimov (approved):

PR #614, opened 2026-08-18
.github: upgrade Actions

sami-nspcc:
Upgrade GitHub Actions to the latest version.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -34,7 +34,7 @@ jobs:
     - uses: actions/checkout@v7
 
     - name: Set up Go
-      uses: actions/setup-go@v6
+      uses: actions/setup-go@v7
       with:
         go-version: '${{ matrix.go }}'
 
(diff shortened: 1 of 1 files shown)


AnnaShaleva (approved):

AnnaShaleva:
Looks like NNS has stale NEF:
```
Run if [[ $(git diff --name-only | grep '^contracts/*' -c) != 0 ]]; then
  if [[ $(git diff --name-only | grep '^contracts/*' -c) != 0 ]]; then
  echo "Fresh version of binaries/manifests should be committed for the following contracts:";
  git diff --name-only;
  exit 1;
  fi
  shell: /usr/bin/bash -e {0}
  env:
    GOTOOLCHAIN: local
Fresh version of binaries/manifests should be committed for the following contracts:
contracts/nns/contract.nef
```

AnnaShaleva:
Something strange, I can't reproduce it locally.

AnnaShaleva:
I re-run the failing `Build contracts and RPC bindings` job, it PASSes now. But it shouldn't have failed.

roman-khimov (approved):

PR #615, opened 2026-08-18
.github: make  bindings/NEF/manifest checkers verbose

AnnaShaleva:
Full diff helps to investigate possible floating compiler bugs.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -90,6 +90,8 @@ jobs:
           if [[ $(git diff --name-only | grep '^rpc/*' -c) != 0 ]]; then
           echo "Fresh version of RPC bindings should be committed for the following contracts:";
           git diff --name-only;
+          echo "Full git diff:";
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -90,6 +90,8 @@ jobs:
           if [[ $(git diff --name-only | grep '^rpc/*' -c) != 0 ]]; then
           echo "Fresh version of RPC bindings should be committed for the following contracts:";
           git diff --name-only;
+          echo "Full git diff:";
+          git diff;
```
roman-khimov, line 94: Then `--name-only` is redundant to me.

AnnaShaleva: I don't think so. Usually it's enough to see the list of files affected, whereas full `git diff` may be large, I don't want to scroll till the end to see if there are some other files affected.

roman-khimov (approved):

